PR #2, opened 2026-06-23
fix(claude-identity): only notify/banner on successful account swap

jdb-117:
## Problem

`_watch.py` calls `notify()` unconditionally immediately after the `switch` subprocess returns, **before** checking its exit code:

```python
r = subprocess.run([self_path, "switch"], ...)
# ...
notify(f"swapped from {email}: {reason}")   # fires even if switch failed
```

On a **single-account** setup, `switch` exits non-zero by design — rotation requires ≥2 accounts (`[ "$total" -ge 2 ] || die "need at least 2 accounts..."`). But `notify()` had already fired, so the operator gets a desktop notification falsely claiming `swapped from <email>` even though no swap happened and the active account is unchanged.

This is the **default starting configuration** (one account, before an overflow is added), so the false notification fires on every trigger for the most common setup.

The in-session banner (the `swap-notification.json` marker) was already correctly gated behind `returncode == 0`, so only the OS notification lied — but it lied consistently.

## Fix

Early-return on non-zero return code with an explicit log line, so both `notify()` and the marker write only run after a **confirmed** swap.

- No behavior change for the ≥2-account path (success still notifies + writes the banner + honors `CLAUDE_AUTOSWAP_RESPAWN`).
- Single-account / failed-swap path now degrades quietly: the attempt is still recorded in the swap-log and the `swap result: rc=...` line; the session simply rides its own cap until the rate-limit window rolls over.

## Test

`python3 -m py_compile hooks/claude-identity/_watch.py` passes. Verified the only diff vs upstream HEAD is this change.

### DIFF
--- a/hooks/claude-identity/_watch.py
+++ b/hooks/claude-identity/_watch.py
@@ -165,8 +165,6 @@ def main(self_path: str) -> None:
             "rc": r.returncode,
         })
 
-        notify(f"swapped from {email}: {reason}")
-
         # After a successful swap: write a swap-notification marker that the
         # statusLine renderer (context-monitor.py) surfaces as a prominent
         # in-session banner ("🔄 swapped j→cc Nm ago — restart pre-swap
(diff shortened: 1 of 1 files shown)


chuycepeda:
Thanks @jdb-117 — you're right, `notify()` fired before the returncode check, so a single-account rotation attempt falsely announced a swap. Fix confirmed. We're folding it into the next release (imported by hand + Co-Authored-By credit — a neighboring in-flight change to `_watch.py` would conflict on merge). I'll link the commit and close here when it ships. 🙏

chuycepeda:
Landed 🎉 — imported in d425ec2 (now on `main`) with your Co-Authored-By credit. The notify + in-session banner now fire only on a *successful* account swap, per your fix — a single-account rotation no longer falsely announces a swap. Thank you, @jdb-117. Closing as imported.

PR #10, opened 2026-07-29
fix(mcps): pin mcp for the three servers that import mcp.server.fastmcp

Pipena:
## The problem

`mcps/setup.sh` installs three bundled servers with an **unpinned** `pip install … mcp`:

| Server | setup.sh line |
|---|---|
| `nano-banana-mcp` | `pip install google-genai mcp -q` |
| `pdf-generator-mcp` | `pip install mcp -q` |
| `spotify-dj-mcp` | `pip install spotipy mcp -q` |

All three do `from mcp.server.fastmcp import FastMCP`. **`mcp` 2.0 removed that module.**

Where the unpinned install resolves to 2.x, the venv builds cleanly, `setup.sh` prints `✓ installed`, and the server then fails at import — surfacing later, elsewhere, as `✘ Failed to connect` in `claude mcp list`.

Existing vaults are unaffected: a venv created before 2.0 keeps its resolution and keeps working. The exposed operators are the ones onboarding, rebuilding a machine, or setting up a second one — the people with the least context to diagnose it.

## Reproduction

Running `setup.sh`'s own command against the repo's own `pdf-generator-mcp/server.py`:

| Command | Resolves to | Result |
|---|---|---|
| `pip install mcp -q` (current) | `2.0.0` | `ModuleNotFoundError: No module named 'mcp.server.fastmcp'` |
| `pip install mcp==1.28.1` | `1.28.1` | imports OK |

Confirmed for all three servers in both directions.

**Scope of verification, stated plainly:** reproduced on macOS with Python 3.12. Whether every platform/interpreter resolves identically has not been tested, so the blast radius is *"wherever pip resolves `mcp` to 2.x"*, not a claim about every machine. The fix is correct either way — pinning a dependency the servers demonstrably require is right independent of how many environments currently break without it.

## The fix

- Three new `requirements.txt` pinning `mcp==1.28.1`, plus `spotipy==2.26.0` / `google-genai==2.8.0` where applicable.
- The three `setup.sh` lines now install `-r requirements.txt`.
- No server code changed.

This brings the three stragglers under the policy `mcps/_index.md` already states — *"Floating `@latest`/unpinned is forbidden: these servers run with live credentials."* The policy was written; these three were never brought under it.

Verified end-to-end: each `.venv` deleted, `setup.sh` re-run, each `server.py` imported.

## Deliberately not fixed

`notebooklm-mcp` and `playwright-mcp` also install unpinned, but neither imports the removed module, so neither is broken today. Widening the diff to them would be speculative rather than tested. They are named in the CHANGELOG entry so the next reader sees them.

## The upstream question — maintainer's call

Two forks I did not decide:

1. **Pin version.** `1.28.1` is the newest 1.x tested here. If you would rather track 2.x, the real fix is porting the three `server.py` files to the new API — a larger change, and yours to scope.
2. **Whether the installer should verify.** The narrow bug is the pins. The broader one is that `setup.sh`'s `✓ installed` is an unconditional `echo` after `pip` — success asserted by position in the script rather than by anything about the artifact, which is why this stayed invisible. A one-line `python -c "import server"` per block would make the class of failure impossible. Not built here because it changes the installer's contract, which is a design call rather than a bug fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,27 @@
 
 ---
 
+## 2026-07-29 — Three bundled MCPs stopped installing, and setup.sh reported success anyway
+
+`hash: TBD`
+
+> **What this delivers.** `mcps/setup.sh` installed `spotify-dj`, `pdf-generator` and `nano-banana` with an **unpinned** `pip install … mcp`. All three import `from mcp.server.fastmcp import FastMCP`, and **`mcp` 2.0 removed that module**. Where that unpinned install resolves to 2.x, it produces a venv that builds cleanly, prints `✓ installed`, and then dies at import — surfacing much later, somewhere else, as `Failed to connect` in `claude mcp list`. Existing vaults are unaffected: a venv created before 2.0 keeps its old resolution and keeps working. **The operators exposed are the ones onboarding, rebuilding a machine, or setting up a second box** — exactly the people with the least context to diagnose it.
+>
+> **Scope of verification, stated plainly:** reproduced on macOS with Python 3.12, where a bare `pip install mcp` today resolves to 2.0.0 and all three servers fail to import. Whether every platform and interpreter resolves the same way has **not** been tested here, so read the blast radius as *"wherever pip resolves `mcp` to 2.x"* rather than as a claim about every machine. The fix is correct either way — pinning a dependency the servers demonstrably require is right independent of how many environments currently break without it.
+
+**What you can now do:**
+- **Set up a new machine and have the MCPs actually run.** Each of the three now ships a `requirements.txt` pinning `mcp==1.28.1` (plus `spotipy==2.26.0` and `google-genai==2.8.0` where they apply), and `setup.sh` installs from it. Verified by deleting each venv, re-running `setup.sh`, and importing each `server.py`.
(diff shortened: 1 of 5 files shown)


chuycepeda:
Landed in #12 (merged as `3990d9bc`) — **this branch's commit `465300b` is in `main`'s history with its original authorship**, not re-implemented.

Closing this PR rather than merging it because the two would have been two pushes to `main`, and every push there raises an "update available" badge on every operator machine. Batching was the whole point; the value is in.

**Why it merged rather than needing a rewrite.** GitHub reported `CONFLICTING`, but `git merge-tree` showed `CHANGELOG.md` was the *only* conflict — `mcps/setup.sh` was untouched since this PR's base and the three `requirements.txt` are new files. So the substance applied clean. The CHANGELOG collision was resolved by re-homing the prose as a `###` subsection of the existing `2026-07-30` entry (this repo keeps one dated entry per day) and replacing `hash: TBD` with the real hashes, now that they exist.

**Verified before taking it:** each block `cd`s into its server dir before `pip install`, so the relative `-r requirements.txt` resolves correctly.

**On the two forks left to the maintainer:**

1. **Pin version** — `1.28.1` taken as-is. Porting the three servers to the 2.x API is the larger change and stays unscoped.
2. **Whether the installer should verify** — the sharper of the two observations, and it generalises past this bug: *"success asserted by position in the script rather than by anything about the artifact."* That is the same shape as a separate bug fixed the same day, where `aios-commit`'s index sync ran `git add --all -- <paths> 2>/dev/null || true` — all-or-nothing, so one non-matching pathspec staged nothing, and the muted error made a repair that never happened indistinguishable from success. Not built here either, but the pattern is now named twice, which is usually the signal it should become a rule rather than a fix.

Also fixed on the way in, reported separately as #11: the same day's licence retrofit moved `LICENSE` but never mentioned `NOTICE`, so a retrofitted repo kept asserting GPL while the entry's own verification passed.

PR #14, opened 2026-08-06
fix: cross-platform portability — vault anchoring, ritual parity, agent probe, Chrome resolution

GregorioRM0:
## Problem class

Three places where the framework assumes macOS or assumes a happy path, and **all three fail silently** — no error, no visible symptom, just a session that quietly isn't what the operator thinks it is. Found while running AIOS daily on Linux Mint; each one is environment-shaped, not vault-shaped, so any operator outside macOS hits them.

---

### 1. Spawned sessions can start outside the vault, so `CLAUDE.md` never loads

`_claude_with_respawn` is called directly from whatever cwd the terminal happens to be in. The `cd` into the vault only exists inside the launcher's macOS AppleScript path, so on Linux (and any plain-terminal spawn) the session starts in `$HOME`, Claude Code finds no `CLAUDE.md`, and **the Session Start Ritual never runs** — no declared context, no observed context, no identity.

The failure mode is what makes this worth fixing: the session opens perfectly normally. The operator believes they're talking to their AIOS and they're talking to a bare Claude.

**Fix:** `cd` into the vault inside the function, so vault-loading is independent of OS, IDE automation, and the caller's cwd.

### 2. The primary-session shorthand never triggers the Session Start Ritual

A fresh `claude` with no bootstrap prompt opens idle and waits. `CLAUDE.md` is loaded, but nothing has taken a turn — so the ritual never runs. `spawn` never had this problem, but only by accident: it *always* passes a bootstrap (`Read <task-file> …`), and that turn is what triggers the ritual. The primary shorthand passing none reads as an oversight rather than a decision, so this makes the two paths behave the same.

`-c/--continue` and `-r/--resume` deliberately keep passing **no** bootstrap: the context is already in the session being resumed, and re-running the ritual would spend a turn re-reading what's already loaded.

### 3. Agent-name probe returns a nonzero exit on the normal case

`CLAUDE.md` instructs a glob probe to find a matching agent file. A bare glob with no match exits 2, which surfaces a red error on **every** spawn whose name doesn't match an agent — and that's the *expected* case for primary sessions, which aren't agents. The practical cost isn't the failed probe; it's that it trains operators to ignore red errors.

**Fix:** `find agents -ipath "*/{name}.md" -print -quit`, which exits 0 on no-match.

### 4. PDF Generator MCP only resolves Chrome on macOS

`server.py` hardcoded the macOS Chrome path, so the MCP fails on Linux and Windows. Now resolves Chrome/Chromium/Edge per-OS at call time, with a `CHROME_PATH` env override for unusual installs.

---

## Testing

Run live in a real vault on Linux Mint 22.3 / bash, Claude Code 2.1.223:

- **(1)** verified against the consumer, not the file: `spawn` from `/tmp`, then asked the new session for its cwd and whether it loaded `CLAUDE.md`. Before: `$HOME`, no ritual. After: vault, ritual runs.
- **(2)** primary shorthand opens and immediately runs the ritual; `-c` / `-r` confirmed still bootstrap-free.
- **(3)** `find agents -ipath "*/does-not-exist.md" -print -quit; echo $?` → `0`, no output.
- **(4)** PDF generated end-to-end on Linux.
- `bash -n` on the wrapper installer, `py_compile` on `server.py`.
- No new arrays introduced, so no bash 3.2 `set -u` empty-array exposure.

## Scope

Deliberately excluded from this PR: the model-default value in `install-wrappers.sh` is a personalization, not a portability bug — upstream should bump its own default on its own schedule, so this branch leaves it untouched.

## The upstream question

(1) and (2) both change *when* the Session Start Ritual runs. I've treated that as a bug — the ritual is documented as the contract for every session and these two paths were skipping it — but if either path skipping the ritual was intentional, say so and I'll narrow the PR to (3) and (4), which are unambiguous.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -41,7 +41,7 @@ If `USER.md` exists, read it. It contains:
 
 1. Your session name is your role (e.g. `accountant`, `lawyer`, `code-review`, `writing`).
 2. **Agent matching** (in order):
-   - **Exact:** glob `agents/**/{name}.md` — first match wins. `agents/custom/{name}.md` always overrides bundled. Wrapper warns on collisions across bundles.
+   - **Exact:** glob `agents/**/{name}.md` — first match wins. `agents/custom/{name}.md` always overrides bundled. Wrapper warns on collisions across bundles. **Probe with a command that exits 0 on no-match** — `find agents -ipath "*/{name}.md" -print -quit 2>/dev/null` (a bare `ls agents/**/{name}.md` exits 2 when nothing matches, surfacing a harmless-but-noisy red error every spawn). A no-match is expected and normal — especially for primary sessions in USER.md, which aren't agents; don't treat the empty result as a failure.
    - **Fuzzy:** no exact file → read `agents/_index.md` (canonical registry) + `agents/custom/_index.md` if present; match session name against agent names/purposes/keywords; pick closest; tell user which matched and why.
    - **No match:** No close agent found → general-purpose worker with session name as role context.
 3. **Greet by name with a contextual message that fits the role** — natural prose, not a fixed line. E.g. `lawyer` → *"Counsel is in, [name]. What do you need me to look at?"* (`[name]` from `about_me.md`; never hardcode). Adapt for any role.
--- a/hooks/claude-identity/install-wrappers.sh
+++ b/hooks/claude-identity/install-wrappers.sh
@@ -146,6 +146,15 @@ _claude_with_respawn() {
   local name="${1:?Usage: _claude_with_respawn <name> [bootstrap] [initial-session-id]}"
   local bootstrap="${2:-}"
   local initial_sid="${3:-}"
+
+  # Anchor every spawned session in the vault so Claude Code auto-loads ~/aios/CLAUDE.md
+  # and runs the Session Start Ritual. The launcher's `cd ~/aios` only covers the
+  # macOS AppleScript path; the in-shell spawn path (and any plain-terminal spawn on
+  # Linux) calls this function directly from whatever cwd the terminal happened to be
+  # in — landing sessions in $HOME with no CLAUDE.md, so nothing loaded. cd'ing here
+  # makes vault-loading independent of OS, IDE automation, and the caller's cwd.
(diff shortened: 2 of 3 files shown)


chuycepeda:
**Merged** — all four, and thank you. This is the first contribution to take portability seriously as a whole rather than one broken path at a time.

**The sharpest catch is the one that looks smallest.** Noticing that the primary shorthand never ran the Session Start Ritual — and then diagnosing *why* `spawn` was fine as *"only by accident: it ALWAYS passes a bootstrap, and that turn is what triggers the ritual"* — is the difference between fixing a symptom and finding a rule. Two paths that should behave identically diverged because one of them accidentally satisfied a precondition nobody had written down. That reasoning went into the merge as much as the code did.

**The `cd ~/aios` in `_claude_with_respawn`** is the one with the widest blast radius: the launcher's `cd` only ever covered the macOS AppleScript path, so every in-shell spawn and every plain-terminal spawn on Linux started wherever the terminal happened to be, with no `CLAUDE.md` in reach — sessions loaded nothing and nothing announced it.

**The `ls` → `find … -print -quit` probe** deserves a mention too. A no-match is the *normal* case for a primary session, and `ls` exits 2 on it — so the framework was printing a red error, every spawn, for the expected outcome.

**One thing I changed on merge, flagged rather than silently fixed:** the PR landed `hooks/claude-identity/install-wrappers.sh` as `100644` — the exec bit was lost. Every documented invocation is `bash <path>`, so nothing actually breaks, but the file shipped `100755` and a mode flip is the kind of regression that surfaces months later on someone invoking it directly. Restored in a follow-up commit (`05f7340`) that says so. Almost certainly a filesystem artifact on your side rather than an edit — worth a `git config core.fileMode` check if you contribute more shell scripts.

**On #15 (multi-account rotation is macOS-only):** staying open deliberately. Four distinct issues in one subsystem is its own PR with its own testing, not something to fold into a batch — and rotation touches Keychain credentials, so a half-ported version is worse than an honestly macOS-only one. If you want to take it, say so and I will not duplicate the work.

Shipped in the `2026-08-11` CHANGELOG entry. 🙏

PR #30, opened 2026-08-14
fix(windows): make the quota autopilot actually run, and stop the installer giving mac-only advice

giuliana-cmd:
# Windows script parity: make the quota autopilot run, and stop the installer giving mac-only advice

Follow-up to the Windows work in `aios-app` — this is the framework half, on the repo you pointed me at.

The starting brief was *"port the remaining bash scripts to PowerShell."* Working through it, that turned out to be the wrong shape for most of them, and this PR argues the narrower version instead. Full reasoning below, because the scope change is the main thing worth reviewing.

## What this does NOT do, on purpose

`install-git-hooks.ps1` already states the framework's doctrine: bash that git invokes *through `sh`* needs no PowerShell rewrite, because Git for Windows always ships bash — only the per-machine install gets a sibling. Applying that consistently:

| Script | Verdict | Why |
|---|---|---|
| `hooks/git/secret-scan.sh` | **not ported** | Invoked by the pre-commit hook (git runs hooks through `sh`) and by `aios-commit`. A second implementation of a *secret scanner* is a drift hazard with a security blast radius. |
| `mcps/setup.sh` | **not ported, fixed instead** | Already runs on Windows under Git Bash and was already debugged there twice (`$PY` probe, `vbin()`). It didn't need a rewrite; it needed its mac-only *advice* corrected. |
| `mcps/atlassian-mcp/run.sh` | **not ported** | An MCP launcher registered via `claude mcp add … -- bash …`. Adding a `.ps1` would contradict the rule above for no gain. |
| `hooks/claude-identity/claude-identity.sh` | **fixed + scheduler added** | The one genuinely Windows-dormant piece. See below. |

Net: 2 new files, 4 modified. If you'd rather have literal PowerShell ports of all four, say so and I'll do it — but I think the table above is the better trade and wanted to make the argument explicitly rather than quietly narrow the scope.

## 1. The identity manager claimed Windows support and could not run at all

`claude-identity.sh` said *"Windows uses the same file layout as Linux"*, and its credential backend genuinely did detect Windows correctly. But **every subcommand failed there** — fail-closed, in three independent ways:

1. **`python3` is the Microsoft Store redirector stub** on a default Windows install: it prints an "install Python" notice and exits non-zero. That killed `current_email()`, so `whoami`/`list`/`switch` never started — and killed `cred_write()`'s pre-write blob validation, a pipeline whose failure trips `die`, so **a swap could never complete**. Fixed with the same interpreter probe `mcps/setup.sh` already carries for the identical stub.
2. **POSIX paths handed to a native interpreter.** Under Git Bash `$HOME` is `/c/Users/you`, and those paths get embedded in Python source run by a *native Windows* Python, which cannot resolve that form. With a real interpreter found, `whoami` still died with `FileNotFoundError: '/c/Users/you/.claude.json'`. Path variables are now normalised once through `cygpath -m` into the mixed form both bash and native Python accept, so every caller is unchanged. `SELF_DIR` is deliberately left alone — MSYS converts it on argv; checked, not assumed.
3. **`grep -c .` exits 1 on zero matches**, and under `set -e` a failing command substitution in an assignment aborts the function. **Not Windows-specific:** any operator who hasn't yet added `## Anthropic accounts` to `USER.md` — the default state of every new install, on every platform — got exit 1 and *no output at all*, instead of the "not listed in USER.md" branch written fifteen lines below for exactly that case.

## 2. Task Scheduler wiring (`install-quota-watch.ps1`)

Third sibling of the launchd plist and the systemd user timer, with a mapping table in the header so a change to one can be mirrored. Three things it had to discover empirically, each documented inline:

- **`Register-ScheduledTask` needs elevation** — `Acceso denegado`, HRESULT `0x80070005` — with or without an explicit `-Principal`. `schtasks.exe /Create /XML` registers the identical definition unelevated. Requiring an admin shell to install a *per-user* quota watcher is the wrong trade.
- **`<LogonTrigger>` is the single element that needs elevation** (bisected element by element: logon-trigger-only fails, repeating-TimeTrigger-only succeeds). So `RunAtLoad` parity is opt-in via `-WithLogonTrigger`, and the default leans on `StartWhenAvailable` with a `StartBoundary` already in the past — systemd's `Persistent=true`, covering the same gap.
- **`[TimeSpan]::MaxValue` serialises to `P99999999DT23H59M59S`** and the scheduler rejects the whole task.

**The first version printed `registered scheduled task` over a registration that had failed**, because the cmdlet surfaced the rejection as a non-terminating error. It now sets the Stop preference and *verifies against the scheduler* afterwards instead of trusting the call — the same lesson as UI buttons that were no-ops while the build stayed green: a success message is not evidence. Relatedly, `schtasks` stderr is redirected *inside `cmd`*, because PowerShell 5.1 wraps a native exe's stderr in an ErrorRecord that the Stop preference would turn into a crash.

## 3. A test, so this is verifiable without risking a real account

`test-identity-cycle.sh` — 25 assertions against a throwaway `CLAUDE_CONFIG_DIR`, ~2s, reads no real credentials. Covers whoami (listed + unlisted), list, two captures, explicit switch, bare rotation, and a malformed blob that must be refused with the live credential untouched. It also asserts the swap does **not** disturb `mcpServers` or `projects` — the property the tool promises and nothing checked.

**Confirmed not a vacuous gate:** 25/25 against the fixed script, **21 failures** against the unmodified one.

## 4. `mcps/setup.sh` — mac-only advice

Two `brew install X` messages fired unconditionally, naming a package manager Windows and most Linux operators don't have. And the Chrome check was a bare `[ -d "/Applications/Google Chrome.app" ]`, so on Windows and Linux it **always** failed and warned "Google Chrome not found" at operators who had Chrome installed — a false warning is worse than none, because it teaches you to ignore the installer's output.

## Verification

Windows 11 Home 26200, Git Bash MINGW64, **non-elevated shell**:

- `test-identity-cycle.sh` → 25/25, and 21 failures against the pre-fix script.
- `whoami` → works interactively **and through `bash -lc`**, which is how a scheduled task invokes it (the login shell has a different PATH — worth testing separately, and it's where I initially mis-verified against an unpatched copy).
- Scheduler: install → `-RunNow` → **scheduler result code 0**, empty task log, one decision line in `quota-watch.log` (`no cache yet — skip`, correct with no statusLine pipe). Lifecycle idempotent: `-Uninstall` on nothing, install, uninstall, `-RunNow` when absent all exit correctly. Task removed after verification; the operator's four pre-existing tasks untouched.
- `bash -n` clean on all three shell scripts; `.ps1` passes the PowerShell parser and is pure ASCII.

## Not verified, and worth stating plainly

- **A live two-account swap on Windows has not been run against real Anthropic credentials** — the operator on this machine has a single account, so the rotation path is covered only by the fixture. `claude-switch` always writes `.claude.json.bak-claude-switch` before touching anything, so the first real rotation is a verification step rather than a leap. Called out in both `SETUP.md` and the autopilot README.
- **`chmod 600` is a no-op on NTFS.** The credential blob is protected by the user profile's ACL and nothing more — which is also exactly how Claude Code itself stores it, so this changes nothing, but the docs no longer imply file-mode protection.

## Docs

The autopilot README's hard precondition instructed Claude to **skip setup entirely** when the platform wasn't Darwin or Linux — which would have left this wiring unused, so that's updated along with `SETUP.md`'s "Windows stays dormant". Both now carry the two caveats above, and the README tells Claude to run the test before trusting the tool with a live account.


chuycepeda:
Merged as `ac2989ff`. Three notes, because this PR shipped more than its title says.

## What I changed on your branch before merging

1. **`KEYCHAIN_SERVICE` became overridable** — `"${AIOS_KEYCHAIN_SERVICE:-Claude Code-credentials}"`. The default is byte-identical, so no keychain entry moves for anyone.
2. **Validation hoisted above the backend `case`**, so it covers both stores, and the error names the actual one — `Keychain 'Claude Code-credentials'` or the file path.
3. **The `PY` probe moved inside the `# ---- credential backend` range**, so `tests/credential-backend.test.sh`'s sed-extraction is self-contained.

## The lesson your PR surfaced, which cost me a real credential

Your suite claimed isolation via `CLAUDE_CONFIG_DIR`. That scopes the **files** — it never scoped the **Keychain**, which is machine-global. I ran the suite on my Mac and section 7, which deliberately writes a malformed blob, wrote straight through to my live Anthropic credential and logged me out. Restored from `~/.claude/identities/`, so no lasting harm. But I had verified the file isolation and *assumed* the Keychain, which is the whole failure in one sentence.

So the suite now **refuses to run — exit 2 — unless the override is in effect**, and CI strips the override from a copy and requires that exit 2. A test that merely *claims* isolation is what caused the incident; this one asserts it.

## And why the 25/25 wasn't 25/25 on macOS

The suite seeded only the file backend. On a Mac that meant three assertions passed by reading my **real** credential, and one couldn't pass at all — a genuine 22/25 reporting as green. It now seeds the platform's actual backend, and there is a **two-OS matrix**, because the backend is chosen by `uname` and a single-OS run certifies exactly one of them. Confirmed not vacuous: 25/25 against the fixed script, **21 failures** against the unmodified one.

## Your scope call was right, and the argument was the contribution

Declining to port `secret-scan.sh` (git runs hooks through `sh`, and a second secret scanner is a drift hazard with a security blast radius), `run.sh`, and `setup.sh` — which needed its **advice** corrected, not a rewrite — is the doctrine `install-git-hooks.ps1` already stated, applied consistently.

I used the same test in today's follow-up (#31): `aios-star-check` got a `.ps1` **only** because its Windows callers are an Electron app and a Node extension host that can't assume Git Bash is on PATH. And because that is still a duplicate, CI now asserts the two implementations agree on their repo allowlist and all four verdicts — the check whose absence is precisely why our two command-bus fulfillers kept drifting.

That PR also proved your Windows lane earns its keep twice over on someone else's code: it caught a `.ps1` carrying 553 non-ASCII bytes with no BOM, and then a `$r:` that PowerShell parses as a **scope qualifier**, which is a parse error — the whole file fails to load. Neither was visible on a machine without PowerShell. There is now a static check on the Ubuntu job for the second class, so nobody has to wait for a Windows runner to learn about it.

## One thing still open, stated plainly

**A live two-account swap on Windows has not been run against real credentials by anyone.** Your machine has a single account, so rotation is covered by the fixture only. `claude-switch` always writes `.claude.json.bak-claude-switch` before touching anything, so the first real rotation is a verification step rather than a leap — but it is unverified, and it says so in `SETUP.md`.

If you ever run that swap on a two-account Windows machine, that result is the most valuable thing anyone could add to this next.

PR #39, opened 2026-08-18
feat(claude-identity): wrapper default model → claude-opus-5[1m]

GregorioRM0:
Opus 5 shipped. The `spawn` / named-session wrappers launch children with an explicit `--model` flag — because `/config` and `/model` are session-scoped and don't propagate to spawned children — and that flag's default was still `claude-opus-4-8[1m]`.

So an operator whose own session had already moved to Opus 5 kept getting the previous generation in **every worker they spawned**. The wrapper is exactly the surface where that stays invisible: the terminal opens, the session greets, the work gets done, and nothing anywhere names the model. `$CLAUDE_MODEL` override unchanged.

### Why this one was measured instead of eyeballed

An invalid or unavailable id passed to `--model` **does not error** — it falls back silently to the session model. That makes *"the spawn worked"* worthless as evidence: a typo'd id and a correct one produce the same working terminal. So the check has to ask what actually served the request, not whether the flag was accepted:

```
$ claude -p --model 'claude-opus-5[1m]' --output-format json
modelUsage: claude-opus-5[1m] → canonicalModel: claude-opus-5, contextWindow: 1000000
```

That is the only check here capable of failing, and it is the one that ran.

### What changed

| File | Change |
|---|---|
| `hooks/claude-identity/install-wrappers.sh` | `model_to_use` default + the two comments that quote it — the examples block, and the `spawn-kill` pgrep note that cites the injected flag |
| `hooks/claude-identity/install-wrappers.ps1` | `$modelToUse` default |
| `plugins/aios/commands/cold-start-interview.md` | identity-table launch command |
| `CHEATSHEET.md` | "Override model" row |

Same four surfaces the `2026-05-28` 4.7 → 4.8 bump named (`8ecf6bb`), so this is a repeat of a known set rather than a fresh enumeration — which is also why I trust the set is complete: `grep -rn 'opus-4-8'` outside `CHANGELOG.md` returns exactly these and nothing else.

There is a side effect on the tier ladder worth naming, because it is the part that isn't obvious from the diff: `--tier judgment` deliberately hard-codes nothing and defers to this default *so that it tracks the lineup*. While the default lagged, `judgment` quietly meant the previous generation while `mechanical` stayed pinned to Sonnet 4.6 — narrowing the very gap the ladder exists to hold open.

### Tested

Linux, bash 5.2 — full `tests/` suite green (16 suites), `install-wrappers.test.sh` 20/20 after the edit.

**Not tested: the PowerShell path.** `.ps1` carries the same one-line default and was mirrored, not executed — there is no Windows machine behind this change, and claiming it because the edit looks identical would be the wrong shape of confidence.

**No new CI lane, on purpose.** Nothing here is a platform assertion, so there is nothing for a lane to make re-checkable. The one fact that could be wrong is the model id, and a test asserting the string would only re-state the change rather than falsify it — the measurement above is what can.

### CHANGELOG

Entry added with `hash: TBD-ON-MERGE` and `2026-08-18` as a placeholder date, following the shape used in #16 — both are yours to set on merge.

---

Disclosure: this branch is agent-authored. The decision to send it, the scope, and the review are mine.


GregorioRM0:
All 14 checks green, including the Windows PowerShell 5.1 lane — so a precision on the "not tested" note above, since a green Windows lane could easily be read as more than it is.

That lane asserts the `.ps1` **parses** under Windows PowerShell 5.1 and stays within the ASCII-only rule. `claude-opus-5[1m]` is pure ASCII, so the edit is inert for both. What it does **not** cover is model resolution — CI has no authenticated Claude CLI, so nothing there can observe which model a Windows `spawn` actually gets.

Net: the untested surface narrows from "the whole PowerShell path" to "a spawn on Windows resolving this id" — which is the same one-line default already measured on Linux, but I have not seen it run there and am not claiming it.

PR #45, opened 2026-08-23
docs(google-workspace): Testing mode is not permanent — it expires the token every 7 days

s1sko7:
## What

`mcps/google-workspace-mcp/personal-account-setup.md` currently tells the reader:

> You do **not** need to publish/verify the app — Testing + test-user is the correct, **permanent** state for a personal setup.

Testing + test-user does authorize correctly, so the setup works on day 1. But Google expires the **refresh token** of an app whose publishing status is **Testing** after 7 days. The state the guide calls permanent has a one-week lifetime.

The symptom is `invalid_grant: Token has been expired or revoked`.

## Why it belongs in the setup guide and not in troubleshooting alone

Traps #1 and #2 in that section fail **during setup** — the operator is in the console, reads the error, fixes it, moves on. This one fails a week later on a machine nobody is watching, and it presents as *"the MCP broke"* rather than *"a policy clock ran out"*. By the time it happens, nobody connects the symptom back to the step that caused it.

That matters more than usual for this particular server, because AIOS itself schedules work on top of it — the daily-ritual pipeline reads Calendar and Tasks through this exact credential store. A weekly outage that requires a human to click a consent screen is a poor foundation for anything automated, and the guide is the only place that can say so in time.

## What the patch does

- Renames the section **two traps → three traps** and adds trap #3.
- Softens trap #2 from *"the correct, permanent state"* to *"enough to authorize and gets you running today — but not a permanent state"*, pointing at #3.
- Adds one troubleshooting row: `invalid_grant` after ~7 days → publishing status still Testing.

## What the patch deliberately does NOT claim

An earlier draft of this change said publishing to Production needs no Google review, reasoning by analogy with the OAuth-login case where basic `email`/`profile`/`openid` scopes are non-sensitive. **That reasoning does not transfer here** — this MCP requests Gmail and Drive, which are exactly the sensitive/restricted scope classes that pull an app into verification. Shipping that sentence would have replaced one confidently-wrong claim with another.

So the patch separates the two things that were being conflated — **publishing status** (what the 7-day clock is attached to) from **verification** (a separate process an app may or may not need) — and then points the reader at the Audience screen as the only current answer, rather than hardcoding a policy detail that Google owns and can change. It also states the alternative plainly: staying in Testing is a legitimate choice, and it costs a manual re-auth every 7 days, forever.

## How it was found

An operator following this guide has been re-authenticating roughly weekly since mid-August, on Windows, with the pipeline going quiet each time until they did. The cause was mis-attributed locally for two weeks before it was traced to the publishing status.

Docs-only change — no code, no behavior.

https://claude.ai/code/session_01AfUDZuZLHAD4kxdDUhiTR1

### DIFF
--- a/mcps/google-workspace-mcp/personal-account-setup.md
+++ b/mcps/google-workspace-mcp/personal-account-setup.md
@@ -4,13 +4,24 @@
 >
 > Written from a real setup (2026-07-22) on a fortress agent machine. Uses the **local** MCP (its own OAuth client, `uvx`-launched) per the AIOS MCP policy — not the claude.ai-hosted Google connector, which is bound to the active claude.ai OAuth grant and breaks on account switch.
 
-## The two traps (read first — they cost the most time)
+## The three traps (read first — they cost the most time)
 
 1. **`Error 403: org_internal` — "can only be used within its organization."**
    The OAuth client belongs to a Google Cloud project whose consent screen is **Internal** (locked to a Workspace org). A **personal gmail can never use an Internal client.** → Create a **fresh project under the personal account itself** (no org → **External** consent available). Do *not* reuse a client from a work/org project.
 
 2. **`Error 403: access_denied` — "has not completed the Google verification process / developer-approved testers."**
-   The app is in **Testing** mode and the account isn't on the test-user list. → Add the account as a **Test user** (below). You do **not** need to publish/verify the app — Testing + test-user is the correct, permanent state for a personal setup.
+   The app is in **Testing** mode and the account isn't on the test-user list. → Add the account as a **Test user** (below). Testing + test-user is enough to *authorize* and gets you running today — but it is **not a permanent state**. See trap #3, which is the bill for stopping here.
+
+3. **The whole thing works, then dies exactly 7 days later — `invalid_grant: Token has been expired or revoked`.**
+   Google expires the **refresh token** of an app whose publishing status is **Testing** after 7 days. Nothing is misconfigured and nothing was revoked by a human: the grant reached its built-in expiry.
+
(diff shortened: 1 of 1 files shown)


chuycepeda:
Merged as `d6fba4c`. Thank you — this is the kind of docs fix that is worth more than most code.

It deletes the word **"permanent"** from a claim with a **7-day** lifetime, and your argument for *where* it belongs is the part I want to keep: traps #1 and #2 fail **during setup**, while the reader is still in the console holding the context. This one fails on a Tuesday three weeks later, on a machine nobody is watching, and presents as *"the MCP broke"* rather than *"a policy clock ran out."* By then nobody connects the symptom to the step that caused it — so the setup guide is the only place that can say it in time. That reasoning is now in the guide.

**The part that decided it for me is what you took OUT.** An earlier draft said Production needs no review, reasoning by analogy from the basic `email`/`profile`/`openid` case — and you cut it, because Gmail and Drive are exactly the sensitive scope classes that pull an app into verification. Replacing a confidently-wrong claim with a second confidently-wrong claim is the most common way a docs fix makes things worse, and you caught your own. Pointing at the Audience screen as *"the only current answer"* rather than hardcoding a policy Google owns is the right call, and it is why this will still be true in six months.

**I checked whether this reproduces here, and it does not — which is the argument for the PR, not against it.** Zero `invalid_grant` across 45,062 lines of this machine's MCP debug log, and `oauth_states.json` — written only during an OAuth consent flow — untouched since **2026-07-28**. Four weeks of continuous operation with no re-auth, so this setup is not in Testing mode and *structurally cannot* produce the failure. Nobody here would ever have found it.

That is the third week running where the most valuable finding came from a platform, locale, or configuration no maintainer runs. It is now a standing request in `CONTRIBUTING.md` for exactly that reason, and your report is a better example of it than the one written there.

One thing worth flagging for your own setup, since you have been re-authenticating weekly: whatever you decide about publishing, the guide now asks you to **write the cost down next to whatever depends on the server**. If a daily ritual reads Calendar through that credential, a weekly silent outage is the actual bug — the consent screen is just where you notice it.

PR #59, opened 2026-08-29
fix: optional close-session no-push for automated wrappers + karpathy-coding frontmatter

medinagarciajose:
## Summary
- `close-session.md`: setting `CLAUDE_CLOSE_SESSION_NO_PUSH=1` before invoking `/close-session` now suppresses its auto-push. Useful for any automated wrapper (a cron, a scheduled agent) that already does its own controlled end-of-run push — a mid-run push from inside `/close-session` would otherwise race that push and land the vault in a half-committed state, out of order with it. Unset (the default) behaves exactly as before, so interactive use is unaffected.
- `karpathy-coding/SKILL.md`: was missing YAML frontmatter entirely (`name`/`description`), so Claude Code registered it empty and the matcher never fired it. Added both fields so the skill actually loads and triggers.

Both fixes have been running without regressions in a private fork.

## Test plan
- [x] `close-session.md` diff verified to apply cleanly against current `main`
- [x] `karpathy-coding/SKILL.md` frontmatter matches the `name`/`description` convention used by other bundled skills

### DIFF
--- a/plugins/aios/commands/close-session.md
+++ b/plugins/aios/commands/close-session.md
@@ -126,9 +126,12 @@ Announce the detected mode: "Detected: **vault session** — writing to daily no
      --note "$HOME/aios/vault/01 - calendar/{YYYY-MM}/{YYYY-MM-DD}.md" \
      --before "## Close of Day" \
      -m "session: {HH:MM} {Topic}" \
-     --block-file /tmp/aios-session-block.$$.md
+     --block-file /tmp/aios-session-block.$$.md \
+     $([ "${CLAUDE_CLOSE_SESSION_NO_PUSH:-}" = "1" ] && echo --no-push)
    ```
    The helper takes a **per-file lock + merge-appends**: under the lock it re-reads the *latest* note and inserts the block before `## Close of Day` (or at the end if that marker is absent — omit `--before` then), then commits the note via `aios-commit`. So N sessions closing at once each land their block **in turn** — ordered, zero clobber, all visible immediately. (Then `rm` the temp block file.)
+
+   **`--no-push` fires when `$CLAUDE_CLOSE_SESSION_NO_PUSH=1` is set.** Set this before invoking `/close-session` from any automated wrapper (a cron, a scheduled agent) that already does its own controlled end-of-run push — a mid-run push from inside `/close-session` would otherwise race that push and land the vault in a half-committed state, out of order with it. Leave it unset for interactive use; the flag defaults to unset everywhere, so every push happens exactly as before unless an operator's own automation opts in.
 6. Update `session-insights.md` (observation buffer — not a log):
    - **Scan existing entries first:**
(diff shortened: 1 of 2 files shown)


chuycepeda:
Merged, both of them — thank you José. Two notes on what we did after, since one of them changed your work and you should hear it from us rather than find it in a diff.

**The race you found is real, and worse than your PR argued.** We checked before merging: `aios-commit` releases its lock *before* pushing (`hooks/aios-commit:336` — deliberately, so a slow network call never makes other sessions wait), so two pushes genuinely race. And a rejected push does **not** self-heal: it prints *"the remote has diverged. Pull/rebase, then push"* and stops. That's exactly right for a human at a terminal and useless inside a cron, where nobody reads it and the vault sits committed-but-unpushed. So this wasn't a hardening — it was an unattended failure with no recovery path.

**We extended it, because `/close-session` pushes from three places and the PR guarded one.** Step 5's `aios-note-append` was covered; step 10's `aios-commit` (Mode A observed context) and Mode B's report commit still pushed. Your wrapper would have hit the race anyway, at a later site, with a symptom that no longer looked connected to the variable you'd set — which is why we finished it rather than merging as-is. **This is not a knock on the work:** the three-site structure isn't visible from the part of the file you were editing. It's now asserted by a test that counts *both* sites and guards and requires them to agree, so a fourth push site added later fails the build instead of silently escaping the flag.

**We renamed the variable to `AIOS_CLOSE_SESSION_NO_PUSH`.** `CLAUDE_*` is Claude Code's own namespace (`CLAUDE_CODE_SESSION_ID`, `CLAUDE_CONFIG_DIR`); framework switches are `AIOS_*` (`AIOS_UPDATE_REINVOKED`, `AIOS_STAR_ASK`). To be fair to your choice, there *is* precedent for framework variables under `CLAUDE_` — `CLAUDE_AUTOSWAP_RESPAWN` is ours — so this was a convention call, not a correction. We did it now specifically because an environment variable becomes a contract the moment someone's automation depends on it, and nothing did yet. **If you have this piloted in your fork, update the name.**

Your expansion form was correct as written and is unchanged. We verified it in bash, zsh *and* sh: unset adds no argument at all, rather than an empty one that would reach the helper as a stray empty pathspec.

**On the skill:** confirmed and it was the only one of 25 bundled `aios` skills without frontmatter — and the H1 reading `# CLAUDE.md` is the tell, it was pasted from a CLAUDE.md and never converted. The symlink into `~/.claude/skills` existed the whole time, so it reported as installed in every check anyone would think to run while being unloadable. Good catch; that class is hard to see precisely because nothing errors.

One process note that's ours, not yours: CI didn't run on this PR until we approved it — GitHub holds workflows on first-time fork contributions. We've approved it, so it's green on the record now rather than only in our local verification. Future PRs from you should run automatically.

Both changes are in `main` and shipped in today's CHANGELOG entry, credited to you.

PR #63, opened 2026-09-04
fix(pipeline-executor): the first account block wins, not the last

gabobarrera00:
## What breaks

`parse_sources()` walks every line of `USER.md` with no notion of which block it is in. For the four single-valued identity fields, the **last** match wins:

```python
for line in content.split("\n"):
    for pattern, key in [
        (r'- Google Tasks list: `(.+?)`', "google_tasks_list"),
        (r'- Google email: `(.+?)`',      "google_email_primary"),
        ...
    ]:
        m = re.match(pattern, line)
        if m:
            config[key] = m.group(1)     # ← overwrites a value already read correctly
```

`### Google accounts` ships as a `*Primary:*` block, and operators routinely add a further one — a destination-only address, a retired account, one explicitly annotated as *not connected via MCP*. Any of those that names its address as `- Google email: \`…\`` silently replaces the primary.

## Why it is expensive

The failure is invisible where it happens and misleading where it surfaces. Every credentials lookup builds its path from this value (`GOOGLE_CREDS_DIR_PRIMARY / f"{email}.json"`), so it resolves to a file that does not exist and `/today` reports:

```
⏭️ tasks: configured but credentials missing — run the setup for this source
```

…while the primary account's OAuth token is valid the entire time. The operator is sent to re-authenticate a connector that was never broken.

That is the **same user-visible line** `tests/pipeline-executor-status.test.sh` was written for ("The operator re-authenticated OAuth five mornings in a row"), reached by a different cause. That fix made the status line name the right precondition; this one removes another way of reaching the wrong state in the first place.

## The fix

First-wins for the four scalar fields — once a key is parsed, a later line cannot replace it.

**Why not a block sentinel.** The obvious fix is to track `*Secondary:*` and skip inside it. That only covers the labels we thought to enumerate, and the second block is operator-written prose that can say anything (`*Old account:*`, `*Not connected:*`, `*Billing only:*`). Ordering is the one thing the template actually guarantees: the primary account is declared first. So the guard is label-agnostic — test 2 covers a block that never says "Secondary".

The `*EXAMPLE ONLY*` lines the template ships are italic (`- *Google email: …*`), so `re.match` never matched them and first-wins cannot hand the example the win. Test 5 pins that.

## Evidence

`tests/pipeline-executor-config.test.sh` loads the **shipped** module and points `SOURCES_PATH` at fixture files, so it exercises `parse_sources()` itself rather than a copy that can drift — the same principle as `source-detector.test.sh`, by import rather than regex extraction, since these values come from the function as a whole.

Against the pre-fix parser:

```
  FAIL 1. canary: a second block does not overwrite the primary email
       expected primary@example.com, got secondary@example.com
  FAIL 1b. canary: nor the primary tasks list
       expected list-primary, got list-secondary
  FAIL 2. any later block loses, whatever the operator called it
       expected primary@example.com, got retired@example.com
7 passed, 3 failed
```

After:

```
10 passed, 0 failed
```

The canary fails against `main`, so the suite cannot pass vacuously. Wired into `validate.yml` next to the source-detector step; the "every `tests/*.sh` is referenced by a workflow" guard passes.

Run live in a real vault before proposing: the value was being overwritten there, and `/today` recovered its Calendar and Tasks futures once the parse was correct.

## Two upstream questions

1. **Should the value loop also route through `_operator_lines()`?** `_mentions_enabled` already does, and the value loop not doing so is an inconsistency. I left it out deliberately — none of these four patterns matches a template line today, so it fixes no currently-reachable class and would widen the diff speculatively. It is defence in depth against a future template that writes a non-italic example. Happy to add it if you want the two scans symmetrical.

2. **No `CHANGELOG.md` entry in this PR.** The checklist asks for one, but every `hash:` must resolve in canonical git and a contributor cannot know the merge hash — and the recent practice (#59) is that the entry lands as a separate maintainer commit afterwards. Say the word and I will add one in whatever form you prefer.

https://claude.ai/code/session_01DhepP9P5QWkS7qGayN9XjN

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -111,6 +111,19 @@ jobs:
           bash tests/source-detector.test.sh
           echo "OK: template prose does not vote on the operator's configuration"
 
+      - name: The first account block wins; a later one cannot overwrite it
+        run: |
+          set -e
+          # parse_sources() walked every line with no notion of which block it was in, so
+          # the LAST match won. "### Google accounts" ships a *Primary:* block and
+          # operators add further ones (a destination-only address, an old account), any
+          # of which silently replaced an address already read correctly. Every
+          # credentials path is built from that value, so /today reported Calendar and
+          # Tasks as "configured but credentials missing" while the OAuth token was valid.
+          # Tests 1, 1b and 2 fail against the pre-fix parser, so this cannot pass vacuously.
+          bash tests/pipeline-executor-config.test.sh
+          echo "OK: a later account block cannot overwrite the primary"
+
       - name: Changelog scan handles multi-hash entries; layer pathspec is an array
         run: |
           set -e
--- a/hooks/pipeline-executor.py
+++ b/hooks/pipeline-executor.py
@@ -237,6 +237,28 @@ def parse_sources():
     if "### Dev projects" in content:
         config["configured"].add("dev-projects")
 
+    # FIRST operator line wins for these identity fields, not the last one.
+    #
+    # These four are single-valued, but the scan below is a plain walk over every line
+    # with no notion of which block it is in — so before this guard the LAST match won.
+    # The template ships "### Google accounts" as a *Primary:* block followed by an
+    # optional second block, and operators routinely add a further one: a
+    # destination-only address, an old account, one explicitly annotated as not
+    # connected via MCP. Any of those that names its address as "- Google email: `…`"
+    # silently overwrote the primary that had already been read correctly.
+    #
+    # The consequence is invisible where it happens and misleading where it surfaces:
+    # every downstream credentials lookup builds its path from this value
+    # (GOOGLE_CREDS_DIR_PRIMARY / f"{email}.json"), so it resolves to a file that does
+    # not exist and /today reports Calendar and Tasks as "configured but credentials
+    # missing" — while the primary account's OAuth token is valid the whole time. The
+    # operator is sent to re-authenticate a connector that was never broken.
+    #
+    # First-wins rather than a block sentinel on purpose: keying on a literal label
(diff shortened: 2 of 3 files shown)


chuycepeda:
Merged — thank you. Two things worth saying back.

**Your diagnosis was right, and it is a second independent cause.** A week ago the same misleading \`/today\` line was fixed from the other end: the status block now names the failing precondition instead of asserting a cause it had not measured. That fix addressed *why the message was unhelpful*. Yours addresses *why the value was wrong in the first place* — a later account block silently overwriting the first. Same symptom, genuinely different root cause, and the two compose rather than overlap. `_scalar_parsed` is the right shape: parse-once is a property of the parser, not something each call site should have to remember.

**Your PR was red through no fault of your own — the defect was ours.** Our `Framework purity` step globbed `<layer>/*/`, reading the **filesystem** rather than git. Your suite is the first CI step in this repo`'s history to execute `hooks/pipeline-executor.py`; Python wrote `hooks/__pycache__/`, and a later step in the same job read that directory as a company namespace. `.gitignore:13` excludes `**/__pycache__/`, so git never saw it — only the glob did. Fixed in #65: the check now asks git what the repo *contains* instead of asking the disk what is *sitting in it*, which makes the whole class impossible. Your branch was rebased onto that and came back 14/14.

So your contribution found two bugs: the one you were aiming at, and a latent one in our own CI that would have hit the next contributor whose test happened to import a hook. That is the more valuable of the two.

One small follow-up we will handle: your comment quotes the string `"configured but credentials missing"`, which the earlier fix removed — it reads `configured but not queried — {reason}` now. Nothing wrong with the code, just a stale reference in a comment, and ours to correct since we moved the string.

PR #83, opened 2026-09-07
feat(statusline): account label reads operator alias from USER.md

fernandezdiegoh:
## What

The `👤` account chip in the statusline (and the `🔄 from→to` swap banner) showed the email's local-part. When an operator has two accounts that share a local-part (same person, two providers), the chip is ambiguous: both render identically, and the only way to tell them apart is the usage percentages.

This reads an optional operator alias from `USER.md` → `## Anthropic accounts`, per numbered line:

```
1. `someone@example.com` — alias **work** · any other prose
```

`{email: alias}` is parsed once per process. No alias on a line → that account keeps the local-part, so existing vaults render exactly as before. `USER_MD_PATH` is honored like in `claude-identity.sh`.

## Why not hardcode

Operator vocabulary stays in `USER.md`; the renderer only knows the convention. Same decoupling the identity wrapper already uses for the account list itself.

## Tested

- Alias resolution for three declared accounts, fallback to local-part for an undeclared one, `?` for empty.
- Full statusline render with a synthetic payload.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1133,6 +1133,16 @@ jobs:
         # only considers .py/.sh.
         run: python3 tests/lint-claude-p.py
 
+      - name: the statusline account chip falls back safely
+        # This renders on EVERY prompt, so a raise in the alias lookup is not a wrong
+        # label — it is a broken statusline. The suite spends most of its effort on
+        # absent/malformed USER.md, and asserts the no-alias case explicitly, because
+        # that is the property keeping every existing vault rendering as before.
+        # Contributed in #83; the self-locate and doc-placement assertions were added
+        # on merge — USER.md is Tier-2 and never syncs, so a convention documented
+        # only there would reach fresh clones and no existing operator.
+        run: bash tests/statusline-account-alias.test.sh
+
       - name: every non-ASCII-printing hook guards Windows stdout
         # Windows consoles are cp1252, so a print() carrying an em dash raises
         # UnicodeEncodeError and the hook does NOT RUN — the failure is absence, not a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,21 @@ This is a **consistency fix, not a new convention** — three hooks already carr
 
 ---
 
+### The statusline account chip can carry a name you choose
+
+`👤` and the `🔄 from→to` swap banner showed the email's **local-part**. Two accounts can legitimately share one — the same person on two providers, or a single Google address behind two separate Anthropic accounts — and then both chips render **identically**, leaving the usage percentages as the only way to tell which account you are on.
+
+Declare an alias per line in `USER.md` → `## Anthropic accounts` and the chip uses it:
+
+```markdown
+1. `you@example.com` — primary
+2. `you@example.com.alt` — overflow · alias **overflow**
+```
+
+**Omit it and nothing changes** — that account keeps its local-part, so every existing vault renders exactly as before. The convention is documented in `hooks/claude-identity/README.md`, which is where the accounts section's format already lives.
+
+*Merged with three additions the contribution could not have anticipated. The default `USER.md` path now derives from the file's own location rather than hardcoding `~/aios` — CI forbids that in `.py` because it depends on the operator's symlink existing, and the shell sibling that does it is exempt only because the rule covers `.py`/`.js`/`.ts`, which is an asymmetry rather than a licence. The convention moved out of the parser into a **Tier-1** doc: `USER.md` is Tier-2 and never syncs, so documenting it only there would have reached fresh clones and no existing operator — a convention living in one parser is folklore. And a suite now covers the fallback path, because this code renders on every prompt, so a raise there is a broken statusline rather than a wrong label.*
+
 ### What you need to do — checks first, then act
 
 > **To the session reading this during `/aios:update`: run the checks and act on what they return.** Both are read-only. Report *"already satisfied — no action"* when a check comes back clean.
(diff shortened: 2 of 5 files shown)


chuycepeda:
Good change, solving a real problem — holding it on two small things, and I want to be precise about which one is yours.

## The CI failure is legitimate, and it is not carelessness

`Migration drift` → *"Framework code must self-locate, not hardcode the ~/aios install path"*. It greps `.py`/`.js`/`.ts` for `expanduser("~/aios`, and your line is:

```python
user_md = os.environ.get("USER_MD_PATH") or os.path.expanduser("~/aios/USER.md")
```

**But your PR body says you honored `USER_MD_PATH` "like in `claude-identity.sh`" — and that is exactly true.** That file, line 158:

```bash
USER_MD="${USER_MD_PATH:-$HOME/aios/USER.md}"
```

You copied the neighbouring file faithfully. The rule simply does not cover `.sh`, so the shell sibling is allowed to hardcode what your Python cannot. **That asymmetry is ours, not a mistake of yours** — and the honest read is that the convention you followed was real.

The fix is one line. `context-monitor.py` sits at `hooks/claude-identity/`, so the repo root is two parents up:

```python
from pathlib import Path
default = Path(__file__).resolve().parents[2] / "USER.md"
user_md = os.environ.get("USER_MD_PATH") or str(default)
```

The rule's reasoning, from the CI error itself: hardcoding depends on the operator's `~/aios` symlink existing, and a `mkdir(parents=True)` on such a path silently creates a phantom vault instead of failing.

## The convention is documented nowhere

This is the one I would not merge without. I grepped canonical for `alias **`:

```
(no matches outside the parser this PR adds)
```

So an operator has **no way to discover the feature exists.** Nothing in `USER.md`, the template, or any doc tells them to write `alias **work**`. The section already documents its own convention in a blockquote — *"Keep backticks around each email or the parser skips it"* — which is exactly the right home for one more sentence.

Worth saying why this matters more than it sounds: the sibling PR in this batch argues that a rule living only in three files is **folklore**. A convention living in exactly one parser is the same shape. The code is fine; the feature is invisible.

## Two smaller notes

- **No test.** Both other PRs in this batch shipped one. `_account_aliases()` is pure given a file path, so it is cheap to cover — including the fallback-to-local-part case, which is the property that keeps existing vaults rendering unchanged.
- **A stale comment removed, correctly.** You dropped `# Green by default — colored red if we couldn't match an Anthropic account`, and it was describing behaviour the code does not have (the return is cyan, and nothing colors red). Good removal — worth mentioning only because silently deleting a comment that was lying is easy to mistake for losing information.

## What I verified works

- `startswith("## Anthropic accounts")` correctly matches a header carrying a parenthetical, which is the real-world shape.
- No alias on a line → local-part fallback, so existing vaults render identically. The backward-compatibility claim holds.
- `try/except` around the whole read means a missing or malformed `USER.md` cannot take down the statusline — right instinct for something that renders on every prompt.
- Memoized once per process, as claimed.

The motivating case is real, incidentally: a vault in this org has two accounts on the **same** Google email with separate Anthropic accounts, so the bare local-part renders them identically and the percentages are the only tell. That is the ambiguity you described, and it is why this is worth landing rather than a nicety.

Send the path fix plus a line in the `## Anthropic accounts` blockquote and I will merge. A test would be welcome; I will not hold the merge on it.

chuycepeda:
Completed it on your branch rather than sending it back — your commit is untouched and still authored by you, mine sits on top. CI is green, merging now.

**First, a correction I owe you.** In my review I said a vault in this org has two accounts sharing a local-part, and offered that as evidence the feature was load-bearing. **That was wrong.** The two accounts are `j@` and `cc@` — distinct, and the chip already renders them differently. What misled me was a line of prose reading *"Same Google email, separate Anthropic account"* — which is about the **Google** account behind them, not the Anthropic login the statusline reads. I checked properly afterwards and the collision does not exist there.

The general case your PR addresses is still real, and the change is still worth landing. But the specific example I cited to justify it was not, and you should not have to discover that later.

## What I added, and why none of it was on you

**1 · Self-locate instead of hardcoding `~/aios`.** This is what CI was failing on, and I want to be clear it was a reasonable read. Your body said you honored `USER_MD_PATH` *"like in `claude-identity.sh`"* — and that file, line 158, does exactly:

```bash
USER_MD="${USER_MD_PATH:-$HOME/aios/USER.md}"
```

You copied the neighbour faithfully. The rule only covers `.py`/`.js`/`.ts`, so the shell sibling is exempt from a constraint your Python is held to. **That asymmetry is ours.** Now:

```python
default_user_md = Path(__file__).resolve().parents[2] / "USER.md"
```

**2 · The convention moved out of the parser.** `alias **x**` was documented nowhere — I grepped canonical and it existed only in the code your PR adds, which means no operator could ever discover the feature. It now lives in `hooks/claude-identity/README.md`, beside the account-line format it extends, and in the `USER.md` template that README writes for new operators.

One placement decision worth knowing, because it is easy to get wrong in the other direction: I deliberately did **not** document it in canonical's `USER.md`. That file is **Tier-2 — `/aios:update` never syncs it** — so a convention documented there reaches fresh clones and *no existing operator*. The component README is Tier-1 and does sync. This is the same argument the Windows PR in this batch makes about a guard living in three files: **a convention living in exactly one place that nobody reads is folklore.**

**3 · A suite, weighted toward the fallback rather than the happy path.** `tests/statusline-account-alias.test.sh`, 11 assertions. The reasoning: this renders on **every prompt**, so a raise in the alias lookup is not a wrong label, it is a broken statusline. Your `try/except` was already the right instinct — the tests just pin it. Most effort goes on absent/malformed `USER.md`, plus two things worth naming:

- **A back-compat control** asserting an account with *no* alias still renders its local-part. Without it, this change could silently relabel every existing vault — the one outcome nobody notices until it confuses them.
- **A section-boundary assertion**, because a numbered list in a later section of `USER.md` would otherwise be read as accounts.

## And one bug of my own, since it is the same class

The CHANGELOG action item I wrote ships a command to detect colliding local-parts. My first version was an unscoped `grep` over the whole `USER.md` — which read numbered lists out of every other section and reported phantom collisions. **That is precisely the bug the section-boundary test I just added asserts against, in the same hour.** It is now `awk`-scoped to its own section and verified against a planted collision so it can still fire.

Your instincts on this were good: memoize once per process, wrap the whole read, fall back rather than fail, and honor `USER_MD_PATH`. The three gaps were framework knowledge you had no way to have — a CI rule that covers Python but not the shell file you copied, and a Tier-1/Tier-2 sync boundary that decides where documentation actually reaches people. Thanks for the fix.

PR #84, opened 2026-09-07
fix(buffer-status): count bullet entries — the framework's two buffer readers disagreed about what an entry is

giuliana-cmd:
## The defect

`hooks/buffer-status.py` splits buffer entries on `^### ` only. But `hooks/route-insight.py` — the tool that **excises** entries from that same `session-insights.md` — resolves **both** `### ` headings and top-level `- ` bullets, with a documented, corpus-measured contract for each:

- indented `  - ` lines are body, not entries
- unindented continuation prose after a bullet legitimately belongs to it (*"measured against the seven live files… 43 bullets are followed by unindented continuation prose that genuinely belongs to them"*)
- a bullet entry ends at the next top-level bullet, any heading level, or an HTML comment

So the framework ships **two readers of one file that disagree about what an entry is**, and the measuring one loses silently.

## What that produced on a live vault

A buffer holding **5 Reinforced anchors and 9 Emerging entries** reported:

```
Emerging   0/10
Reinforced 0/5
  reading the buffer in full costs ~0 tokens

Within contract — nothing to dispose.
```

exit `0`. The buffer was in fact **at** the Reinforced cap, one off the Emerging cap, and ~7,900 tokens to read in full.

Every guard upstream of the split passes on that file — non-empty, `## ` sections present, both stages named — which is why nothing caught it. And `main()` annotates its own ValueError path *"Loud, and exit 2. Never a healthy-looking zero."* This is that exact defect class occurring inside the tool built to reduce it.

Worth noting for triage: the tool is called by `/close-day`, so on any bullet-form buffer the close ritual has been reading a clean bill of health.

## The fix

`_entry_chunks()` resolves heading style first, then bullet, **mirroring `route-insight.py`'s boundary rules** so the two tools cannot drift apart again. Entries are no longer produced by a regex split on the delimiter, so a `class:` line sitting on a bullet's indented continuation parses exactly like one under a heading.

A backstop is kept for the case the original code got wrong for a different reason: a stage holding substantive prose that matches **neither** style now refuses (exit 2) rather than reporting zero. Substance is measured outside HTML comments, so a stage holding only `<!-- ROUTED … -->` tombstones stays legitimately empty and still exits 0.

## Tests

`23 → 30 passed, 0 failed`. Seven assertions in `tests/buffer-status.test.sh`:

- bullet entries counted in both stages
- indented facets, blockquote facets and tombstones **not** counted
- `class:` fields parse inside a bullet entry
- **control** — `### ` heading entries still counted (without it, the change could swap one blind spot for another)
- **control** — a tombstone-only stage still measures clean at exit 0
- an unknown entry shape refuses with exit 2

## Not addressed here

While running this suite on Windows: **5 assertions fail regardless of this change** (`upstream/main` scores 18 passed / 5 failed) because cp1252 stdout cannot encode the em dashes and `⚠` in the report. Three hooks already guard this with `sys.stdout.reconfigure`; `buffer-status.py` does not. Filed as a separate PR — it touches this file's import block, a different region, so merge order does not matter.

https://claude.ai/code/session_01LSYsCZsWA8hA6VxHHJUiXu

### DIFF
--- a/hooks/buffer-status.py
+++ b/hooks/buffer-status.py
@@ -83,6 +83,51 @@ LEGACY_ROUTE = re.compile(r"^\s*\*\*Route to:\*\*\s*(.+)$", re.M | re.I)
 VALID_CLASSES = {"behavioural", "method"}
 
 
+# An entry is a `### ` heading OR a top-level `- ` bullet. This is not a new
+# convention: `route-insight.py` — the tool that EXCISES entries from this same
+# file — resolves both styles and documents why (indented `  - ` lines are body,
+# not entries; unindented continuation prose after a bullet legitimately belongs
+# to it, measured at 43 live instances). Counting only `### ` here meant the two
+# tools disagreed about what an entry is, and the reader silently lost.
+TOP_BULLET = re.compile(r"^- \S")
+ANY_HEADING = re.compile(r"^#{1,6} ")
+
+
+def _entry_chunks(body):
+    """Entry text blocks of ONE style, heading first then bullet.
+
+    Boundaries are style-dependent, exactly as in `route-insight.py`: a bullet
+    entry ends at the next top-level bullet, ANY heading level, or an HTML
+    comment (the `<!-- ROUTED ... -->` tombstones are file furniture, never
+    entries), so its indented facets travel with it.
+    """
+    lines = body.split('\n')
+    starts = [i for i, l in enumerate(lines) if l.startswith("### ")]
+    style = "heading"
+    if not starts:
+        style = "bullet"
+        starts = [i for i, l in enumerate(lines) if TOP_BULLET.match(l)]
+    chunks = []
+    for start in starts:
+        end = len(lines)
+        for j in range(start + 1, len(lines)):
+            ln = lines[j]
+            if ANY_HEADING.match(ln):
+                end = j
+                break
+            if style == "bullet" and (TOP_BULLET.match(ln) or ln.lstrip().startswith("<!--")):
+                end = j
+                break
+        chunks.append('\n'.join(lines[start:end]))
+    return chunks
+
+
+# A section body is "substantive" if it carries prose outside HTML comments — a
+# stage holding only routed-entry tombstones is legitimately empty.
+def _has_substance(body):
+    return len(re.sub(r"<!--.*?-->", "", body, flags=re.S).strip()) >= 80
+
+
 def parse(text):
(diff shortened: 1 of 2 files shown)


chuycepeda:
Merged. I reproduced this on `main` before reading your fix, and it is exactly as you described:

```
Emerging   0/10
Reinforced 0/5
  reading the buffer in full costs ~0 tokens

Within contract — nothing to dispose.        exit 0
```

on a fixture holding **5 real entries**. Two Reinforced, three Emerging, one with unindented continuation prose. Nothing about that output hints at a parse failure.

**The diagnosis is the valuable half, and it is airtight.** `route-insight.py` documents the bullet contract at line 14 and resolves both styles; `buffer-status.py` split on headings only. Two readers of one file, disagreeing about what an entry is, and *the measuring one loses silently while the excising one is correct* — which is the worst possible split, because the tool that reports is the one you would consult to find out whether anything is wrong.

I wrote that tool, and the part I keep looking at is that `main()` already carried the line *"Loud, and exit 2. Never a healthy-looking zero."* The promise was in the file. The parse below it could not keep it, because the guards upstream of the split all pass on a bullet-form buffer — non-empty, `## ` sections present, both stages named. **A precondition check cannot catch a wrong answer that its own preconditions permit**, and I did not think past the preconditions I had written.

Your `/close-day` note is the one that actually matters for triage, and it is why this got a CHANGELOG action item that tells the session to **report explicitly if the numbers changed** rather than just re-run the tool — an operator whose close ritual has been saying "within contract" needs to know the zero was the bug, not the state.

## What I checked

- **Ground truth match** — your fix reports 3/10 and 2/5 on my fixture, and parses `class:` out of a bullet's indented continuation.
- **The regression that mattered most** — a live heading-form buffer reads byte-identical before and after (10/10, 5/5). The change could have swapped one blind spot for another and your heading control is what rules that out.
- **Both backstops** — tombstone-only stage still exits 0; unknown shape refuses with exit 2.

One note on my own review, since it is instructive: my first probe of the exit-2 path got exit `0` and I briefly had it down as a discrepancy against your PR body. The fixture prose was **74 characters** against your 80-char substance threshold. Your code was right and my test was too short — which is a small demonstration of why the threshold needs to exist at all, and why measuring substance outside HTML comments is the correct place to draw it.

`30/30` locally, `14/14` in CI after a rebase onto the Windows fix. Both landed.

PR #85, opened 2026-09-07
fix(hooks): guard Windows stdout in every hook that prints non-ASCII, plus a lint

giuliana-cmd:
## The defect

Windows consoles default to **cp1252**. A `print()` carrying an em dash, an arrow or an emoji raises `UnicodeEncodeError` there — so the failure is not a mangled glyph, it is **the hook not running**. The framework's prose style uses em dashes liberally, which makes this a standing trap rather than an edge case.

Measured on Windows with `tests/buffer-status.test.sh`:

| | result |
|---|---|
| `upstream/main` on Windows | **18 passed, 5 failed** |
| same tree, `PYTHONIOENCODING=utf-8` | **23 passed, 0 failed** |

All five failures are the same crash inside the tool's own report line — no logic defect. A Windows operator cannot run the framework's own suite, and any hook in this set dies mid-report rather than degrading.

## Why this is a consistency fix, not a new convention

The guard **already exists** in three hooks and was simply applied inconsistently:

- `hooks/pipeline-executor.py` — *"Fix Windows encoding — cp1252 doesn't support emojis in output"*
- `hooks/route-insight.py`
- `hooks/claude-identity/context-monitor.py`

Six hooks that print non-ASCII did not have it: `buffer-status`, `claude-identity/_resume`, `markitdown-convert`, `openrouter`, `transcribe`, `video-watch`. Same idiom, placed after the imports.

## The lint

A rule living in three files is folklore, and **CI runs on ubuntu, so this class is invisible here by construction** — the lint is the only thing standing between it and every Windows operator.

`tests/lint-windows-stdout.py` flags any hook under `hooks/` (excluding `custom/`) that emits non-ASCII without the guard, and prints the exact form to add. Wired into the `primitives` job, so it inherits that job's `if: github.event.repository.fork != true` — runs on contributor PRs, skips in an operator fork.

**Verified by mutation, not by observing it pass.** Removing the guard from one hook turns it red and names the file and offending lines; restoring it returns green.

That mutation also exposed the linter **reproducing the very bug it detects** — it echoes the offending source lines, which are non-ASCII by definition, so it crashed on cp1252 *while reporting the defect*. It now carries the guard it enforces.

## Note

This touches `hooks/buffer-status.py`, as does the bullet-entry fix in the companion PR. The two edits sit in different regions (import block vs. `parse()`), so they are independent and merge order does not matter.

https://claude.ai/code/session_01LSYsCZsWA8hA6VxHHJUiXu

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1133,6 +1133,19 @@ jobs:
         # only considers .py/.sh.
         run: python3 tests/lint-claude-p.py
 
+      - name: every non-ASCII-printing hook guards Windows stdout
+        # Windows consoles are cp1252, so a print() carrying an em dash raises
+        # UnicodeEncodeError and the hook does NOT RUN — the failure is absence, not a
+        # mangled glyph. Measured: hooks/buffer-status.py scored 18 passed / 5 failed on
+        # Windows and 23 / 0 elsewhere, entirely from its own report line.
+        #
+        # The convention already existed in three hooks (pipeline-executor.py,
+        # route-insight.py, claude-identity/context-monitor.py) and was applied
+        # inconsistently to the rest, which is the shape a rule takes when nothing checks
+        # it. CI runs on ubuntu, so this class is invisible here by construction — the lint
+        # is the only thing standing between it and every Windows operator.
+        run: python3 tests/lint-windows-stdout.py
+
       - name: every tracked file passes the secret scanner
         # THE GAP THIS CLOSES. tests/secret-scan.test.sh proves the scanner behaves
         # correctly on FIXTURES. Nothing ran it across the tracked tree — so canonical
--- a/hooks/buffer-status.py
+++ b/hooks/buffer-status.py
@@ -51,6 +51,16 @@ import re
 import sys
 from datetime import date, datetime
 
+# Windows consoles default to cp1252, which cannot encode the em dashes and
+# symbols this script prints — so on Windows the report raised
+# UnicodeEncodeError and aborted instead of printing. Already guarded this way
+# in pipeline-executor.py, route-insight.py and claude-identity/context-monitor.py;
+# these callers were simply missed. Safe on macOS/Linux, where stdout is UTF-8.
+if sys.platform == "win32":
+    sys.stdout.reconfigure(encoding="utf-8")
+    sys.stderr.reconfigure(encoding="utf-8")
+
+
 DEFAULT_PATH = os.path.expanduser(
     "~/aios/vault/00 - notes/context/observed/session-insights.md"
 )
--- a/hooks/claude-identity/_resume.py
+++ b/hooks/claude-identity/_resume.py
@@ -78,6 +78,16 @@ import json
 import os
 import subprocess
 import sys
+
+# Windows consoles default to cp1252, which cannot encode the em dashes and
+# symbols this script prints — so on Windows the report raised
+# UnicodeEncodeError and aborted instead of printing. Already guarded this way
+# in pipeline-executor.py, route-insight.py and claude-identity/context-monitor.py;
+# these callers were simply missed. Safe on macOS/Linux, where stdout is UTF-8.
+if sys.platform == "win32":
+    sys.stdout.reconfigure(encoding="utf-8")
+    sys.stderr.reconfigure(encoding="utf-8")
+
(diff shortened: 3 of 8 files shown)


chuycepeda:
Merged. I verified every claim rather than taking the body on trust, and all of them hold.

- The three hooks you named (`pipeline-executor`, `route-insight`, `claude-identity/context-monitor`) are exactly the three that already carried it, so **"consistency fix, not new convention"** is the right framing and it is what makes this uncontroversial.
- The mutation genuinely turns the lint red and names the offending file and lines.
- It is wired into `primitives`, so it inherits that job's `if: github.event.repository.fork != true` — runs on contributor PRs, silent in an operator fork. Correct placement.

**The line that made this worth more than the fix:** *"CI runs on ubuntu, so this class is invisible here by construction."* That is the actual argument for the lint, and it generalises past this bug — a guard whose failure mode cannot occur on the runner needs a check that reasons about the code rather than executing it. And the linter reproducing the very bug it detects, found by mutating a hook to see the red line, is the kind of detail that only shows up when someone verifies red before trusting green.

## One follow-up, non-blocking

Detection is **line-local** — `print(` and a non-ASCII byte on the *same* line. So this shape passes and still crashes:

```python
BANNER = "✓ done"   # non-ASCII lives here
print(BANNER)       # this line is pure ASCII
```

I planted that in a throwaway hook and got a real `UnicodeEncodeError: 'charmap' codec can't encode character '✓'` under `PYTHONIOENCODING=cp1252`, while your linter reported clean.

**Not a blocker, deliberately.** After this merge every hook carries the guard, so the gap cannot bite anything that exists today — only a *future* hook that routes non-ASCII through a constant or an f-string variable. Whether it is worth closing is a real trade: catching indirection means either an AST pass or accepting false positives on any file that merely *contains* non-ASCII, and a linter that cries wolf is one people learn to skip. Your narrow rule is the right default; I would rather it be documented than widened.

If you want it, the cheap 80% is to also flag a module-level assignment of a non-ASCII string literal in a file that prints at all. Your call — and if you would rather leave it, say so and I will note the boundary in the docstring instead so the next reader knows it is a choice rather than an oversight.

Also worth saying: this landed alongside your buffer PR touching the same file, and your note that the two sit in different regions was correct — I checked the hunks (lines 51-67 vs 73+), no overlap, merge order genuinely did not matter. Sequencing your own PRs and saying why is not a small thing.

giuliana-cmd:
**(b)** — leave the rule narrow, document the boundary.

Your own argument decides it: the gap cannot bite anything that exists today, and an AST pass — or false positives on any file that merely *contains* non-ASCII — is real complexity bought for a hypothetical. A linter people learn to skip is worse than a narrow one.

Two things worth putting in the docstring alongside the boundary, so the next reader gets the trigger and not only the limitation:

1. **What would make widening worth it:** a new hook that routes non-ASCII through a constant or an f-string variable. Until one exists, the blind spot has nothing to be blind to — which is the actual reason to wait, rather than "it seemed unlikely".
2. **Your repro documents the gap better than prose would** — the `BANNER` / `print(BANNER)` pair, with the note that it raises a real `UnicodeEncodeError` under `PYTHONIOENCODING=cp1252` while the linter reports clean. A reader who sees the failing shape will recognise it in their own diff; a reader who sees "detection is line-local" will not.

## One option neither of us listed

If you ever want the gap closed *without* the AST pass: **invert the predicate.** Instead of detecting non-ASCII near a `print(`, require the guard in **any hook that prints at all**.

That stays line-local and grep-cheap, has **zero** false positives — it never inspects the string, so it cannot be wrong about one — and closes indirection by construction, because it stops asking *what* gets printed. The cost is the two-line guard in hooks that only ever emit ASCII, which is a no-op on every platform but Windows.

I am not proposing it for this PR. I am noting it so the docstring can record that the narrow rule is a choice with a known cheap exit, rather than a dead end that needs an AST pass to escape.

## On the 74-vs-80 characters

That is the detail I would have wanted from a reviewer and did not expect to get. You had your own probe down as a discrepancy against my PR body *before* you concluded your fixture was short — and the order matters, because the other order is exactly where a reviewer talks an author out of a correct threshold. Worth saying out loud, since the same instinct is what made you reproduce both bugs on `main` before reading either fix.

— Sol (Giuliana's AI)

PR #89, opened 2026-09-07
fix(buffer-status): report the file's cost, not only the entries'

matiasmacera:
## What was wrong

`hooks/buffer-status.py` summed the characters of the **entries** and printed the result as *"reading the buffer in full costs ~N tokens"*. A session does not read the entries — it reads the **file**, which also carries every `<!-- ROUTED … -->` tombstone, the front matter and any prose outside an entry. On a buffer whose routed-entry receipts had accumulated, the report announced a cost under half of what every load actually paid, and printed *"Within contract — nothing to dispose"* over the share it happened to measure.

Same class as #84, one step removed: not a reader that **missed** entries, but one that measured the right thing and **titled it as a different thing**. The caps govern the entries; the token bill is the file; naming one as the other hides exactly the share the caps cannot see.

## What changes

- The report prints both, named for what they are:
  ```
  entries      ~N tokens   (what the caps govern)
  WHOLE FILE   ~M tokens   (what every session actually pays)
  ```
  and a warning with the percentage when the file is ≥ 25 % furniture.
- JSON callers get `entry_chars`, `approx_tokens_entries`, `file_chars`, `approx_tokens_file`. **`approx_tokens_to_read_in_full` keeps its value and meaning** — grepped the framework, no consumer outside the hook, kept anyway so nothing drifts.
- `hooks/_index.md` row updated; CHANGELOG section added under the 2026-09-07 entry (no operator action required — the next `/close-day` simply shows the second line).

## Evidence

- `tests/buffer-status.test.sh`: 4 new assertions — a furniture-heavy fixture (40 tombstones), a lean **control** that must *not* warn, the warning text, and the original key's stability. **34 passed, 0 failed.**
- `tests/lint-windows-stdout.py`: clean (the new print lines carry an em dash and `⚠`; the guard from #85 covers them).

## Checklist (CONTRIBUTING → Before-you-open-a-PR)

- [x] Right home: canonical hook.
- [x] Zero operator data in the diff (grepped for names, paths, dates, figures).
- [x] Index updated (`hooks/_index.md`).
- [x] Not a command — no 3-location sync.
- [x] Not a shared-state write — the hook is read-only (its own test pins byte-identity).
- [x] CHANGELOG entry added.


chuycepeda:
Merged, after fixing one thing on your branch — and it is worth spelling out because it is a defect class this repo has now been bitten by twice today.

## Your commit deleted 25 CHANGELOG lines, and git reported no conflict

The diff read `CHANGELOG.md +6/-25`. Those 25 removed lines were another PR's subsection (#83, the statusline alias), its action item, and two hashes from the entry's `hash:` field. I verified with `git merge-tree --write-tree` rather than guessing:

```
merge-tree exit: 0        ← clean, NO conflict
#83 subsection present: 0 ← gone
action item 3 present:  0 ← gone
```

**The loss was silent.** Git found nothing to complain about, so nothing would have surfaced it at merge time.

**And I was wrong about the cause on first read.** My assumption was a stale branch base — it is not: your base was `ecca3bd`, current `main`, zero behind. What almost certainly happened is that you wrote the entry against a copy of `CHANGELOG.md` read earlier in your session, before #83 merged at 08:45, then committed the whole file. **That is not a mistake anyone would catch by re-reading their own diff**, because the diff looks like exactly what you intended plus some noise.

For what it is worth, the framework has a memory note for the cousin of this — command files snapshotted at session start serving a stale copy after an update. Same shape, different file. The practical guard on a day like today, when `main` moves every hour: re-read `CHANGELOG.md` from disk immediately before writing the entry, not from whatever the session loaded.

I resolved it additively — kept every subsection, rejected every empty-side deletion, and restored the missing action item verbatim from `main`. Your diff is now `+51/-2`, all of it yours.

## On the finding itself

This is the sharpest of the three, and the framing is why: ***"not a reader that missed entries, but one that measured the right thing and titled it as a different thing."*** That is a distinct failure from #84 and a harder one to see, because the number was never wrong — only its label. Verified on a furniture-heavy fixture:

```
entries      ~3 tokens     (what the caps govern)
WHOLE FILE   ~458 tokens   (what every session actually pays)
⚠ 99% of the file is NOT entries — comments, tombstones, front matter.
```

**And the line that decides it:** *"the caps govern the entries; the token bill is the file; naming one as the other hides exactly the share the caps cannot see."* A buffer can sit comfortably inside `10/10` while every session pays for a tombstone graveyard, and the report that was supposed to catch that was measuring the wrong denominator.

Keeping `approx_tokens_to_read_in_full` at its old value and meaning — after grepping for consumers and finding none — is the right instinct. A key that changes meaning silently is worse than a new key.

34/34 on your suite with the furniture fixture and its lean control, 14/14 in CI, and the whole-file measurement is now the number `/close-day` reports.

PR #91, opened 2026-09-07
feat(spawn): --profile <name> loads only the MCP servers a worker needs

matiasmacera:
## What it adds

`spawn --profile <name> <worker> "<task>"` starts the worker with `--strict-mcp-config --mcp-config ~/.aios/mcp-profiles/<name>.json` — only the servers that file declares. Omit the flag: byte-identical behaviour to today.

Same plumbing as `--tier` / `--model`: `spawn()` validates and sets `spawn_profile`, the launcher exports `CLAUDE_MCP_PROFILE`, `_claude_with_respawn` resolves it to `mcp_args`. An unknown profile **refuses inside `spawn()`** and lists what exists — validating in the launcher would open a worker with the full default MCP set, which looks exactly like the flag being ignored.

## What it is for (and not for)

Not tokens: measured, a profile trims 2–4k off a ~50k startup because MCP tools load deferred. It is for blast radius (a file sweep should not hold Gmail/Drive/home-automation credentials) and for churn — a server that disconnects mid-session invalidates the whole prefix and pays a full cache rebuild regardless of size. CLAUDE.md § Spawning carries that, plus two findings that cost an hour each when met blind: per-project `.mcp.json` keys off the directory (useless on a vault-anchored setup), and OAuth http servers renamed in a profile load but cannot be called until authorised once under the new name.

## Evidence

- `tests/install-wrappers.test.sh`: +6 assertions — unknown profile → non-zero, names the flag, writes no launcher; and the mechanism is wired end to end. **26 passed, 0 failed.**
- Live: verified with `claude -p` that a profile-scoped session lists exactly the declared servers and that a stdio one answers a real call; the OAuth caveat was found the same way.

## Checklist

- [x] Canonical home (wrapper + CLAUDE.md). · [x] Zero operator data (grepped). · [x] Not a command — no 3-location sync. · [x] Not shared-state — no 2-actor test. · [x] CHANGELOG section under 2026-09-07 (stacks on #89's; trivial merge).


chuycepeda:
Merged. Verified the mechanism end to end and the two decisions I would have gotten wrong.

**Validating inside `spawn()` rather than the launcher is the right call, and the reason you give is the whole argument:** launcher-side validation *"would open a worker with the full default MCP set, which looks exactly like the flag being ignored."* A guard that fails toward the permissive state, silently, is worse than no guard — the operator believes they scoped the worker and nothing says otherwise.

**And thank you for being honest about what it is not.** *"Not tokens: measured, a profile trims 2-4k off a ~50k startup because MCP tools load deferred."* That number is what makes the real case credible: this is **blast radius**, not economy. A file sweep should not hold live Gmail/Drive credentials, and that is rung 3 of the containment ladder arriving as a mechanism rather than advice. The churn point is the sharper half — a server disconnecting mid-session invalidates the whole prefix and pays a full cache rebuild regardless of size — and I had not thought about profile scope in those terms at all.

The two findings you folded into `CLAUDE.md` are the kind that only come from doing it: per-project `.mcp.json` keying off the directory (useless on a vault-anchored setup, where every session runs from the same root) and an OAuth http server renamed in a profile loading but not being callable until authorised once under the new name. Both cost an hour blind; neither is inferable from the flag's documentation.

One note on sequencing, since you flagged the stack: I merged this **before** #89, because #89 needs a CHANGELOG fix first — see that thread. #92 follows. Nothing in yours conflicted.

PR #92, opened 2026-09-07
feat(hooks): pre-push refuses off-limits remote owners — inert until configured

matiasmacera:
## What it adds

`hooks/git/pre-push`: refuses a push whose remote **owner** is on a per-machine list — both URL shapes, case-insensitive, whole-segment match. The push is the one place the intent is unambiguous; a permission glob over `gh`/`git` command lines is best-effort by construction, this is not.

**Generic by design.** The list lives in `git config --global aios.blockedRemoteOwners "…"`, never in the file, so the framework ships with no operator's organisations inside and a fresh clone is inert (exit 0) until the operator names one. The escape hatch is explicit and prints that it was used: `AIOS_ALLOW_BLOCKED_REMOTE=1 git push …`. `install-git-hooks.sh` now says how to configure it; `core.hooksPath` already makes it active wherever the hooks are installed.

## Evidence

`tests/pre-push.test.sh` — **11 passed, 0 failed**: inert with no list; https + ssh blocked; uppercase owner blocked; unrelated owners pass; `AcmeCorpOSS` passes (no prefix leak); escape hatch allows and announces; refusal names the key and the matched owner. Runs against a scratch `GIT_CONFIG_GLOBAL`, never the machine's real config. bash 3.2-compatible (no arrays beyond what the hook itself avoids, no `mapfile`).

## Checklist

- [x] Canonical home (`hooks/git/`). · [x] Zero operator data — the blocked names are `AcmeCorp`/`anotherorg` fixtures. · [x] Not a command. · [x] Shared-state: read-only guard, exercised by its own test with output above. · [x] CHANGELOG section under 2026-09-07 (stacks on #89's; trivial merge).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -23,6 +23,7 @@
 # the next addition cannot slip through even if it doesn't match these patterns.
 hooks/aios-*            text eol=lf
 hooks/git/pre-commit    text eol=lf
+hooks/git/pre-push      text eol=lf
 
 # `hooks/git/pre-commit` above is the one with teeth: git invokes hooks through `sh`, so a CRLF
 # checkout does not merely warn — the hook fails to execute, and the secret scan that gates every
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1133,6 +1133,14 @@ jobs:
         # only considers .py/.sh.
         run: python3 tests/lint-claude-p.py
 
+      - name: the pre-push owner guard is inert until configured
+        # Contributed in #92. The property that matters is INERTNESS: canonical ships
+        # with no organisation names inside it, and a fresh clone must exit 0 until the
+        # operator names one in `git config`. A guard that blocks by default would be
+        # discovered as a broken push on someone else's machine. The suite runs against
+        # a scratch GIT_CONFIG_GLOBAL so it can never read or write the real config.
+        run: bash tests/pre-push.test.sh
+
       - name: the statusline account chip falls back safely
         # This renders on EVERY prompt, so a raise in the alias lookup is not a wrong
         # label — it is a broken statusline. The suite spends most of its effort on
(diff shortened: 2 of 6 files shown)


chuycepeda:
Merged. Two CI failures were framework knowledge you could not have had, so I fixed them on your branch rather than sending it back — and split them out of your commit, because `rebase --continue` had folded them in and that would have credited you with a fix you did not write.

**1 · `.gitattributes` needs a rule per tracked executable.** `validate.yml` fails the build without one, deliberately: git invokes hooks through `sh`, so a CRLF checkout breaks the shebang on Windows. Your file's own sibling carries exactly this rule two lines up —

```
hooks/git/pre-commit    text eol=lf
hooks/git/pre-push      text eol=lf   ← added
```

**2 · A committed `tests/*.sh` must be invoked by some workflow.** A derived guard fails otherwise, and it exists because *a suite nobody runs is worse than no suite — it reads as coverage while measuring nothing.* Wired into the same job as its siblings.

Same category as a gap that caught another PR this morning: our rule covers one file type and not the neighbour a contributor would reasonably copy from.

**Also resolved a CHANGELOG conflict against #91 by keeping BOTH subsections.** Git offered them as alternatives and they are not — that shape is worth watching, because it resolves clean with no conflict marker in one direction and silently drops a section.

## On the design

**The inertness is the right call and the reason is the one you gave:** *"the framework ships with no operator's organisations inside and a fresh clone is inert (exit 0) until the operator names one."* A guard that blocked by default would be discovered as a broken push on someone else's machine, which is the worst possible place to learn a rule exists. I put that reasoning into the CI step's comment so the next person to touch the suite knows which property to protect.

**And the push being the enforcement point is the sharp observation:** *"a permission glob over `gh`/`git` command lines is best-effort by construction, this is not."* That is true and it generalises — a guard on the *intent* is guesswork; a guard on the *irreversible act* is a guarantee. The escape hatch announcing itself is the other half: `AIOS_ALLOW_BLOCKED_REMOTE=1` that prints it was used means a bypass leaves a trace rather than being indistinguishable from no rule.

11/11 on your suite, 14/14 in CI, and the test running against a scratch `GIT_CONFIG_GLOBAL` rather than the machine's real config is the detail that made it safe to run here without thinking twice.

Issue #93, opened 2026-09-07
CLAUDE.md is ~19k tokens per session — proposal to condense it rule-for-rule, verified by behaviour (discussion before PR)


matiasmacera:
## Proposal (discussion before code — this is architectural)

Condense `CLAUDE.md` by roughly 40 %, rule for rule, with no rule removed — the way an operator vault condensed its own observed files this week.

## Why it is worth a discussion rather than a surprise PR

`CLAUDE.md` is the behavioural contract, loaded into every session. Measured with `claude -p … --output-format json` from a directory holding only `CLAUDE.md`: **~19,000 tokens** (72 KB, 534 lines) — the second-largest single item of a session start after the operator's own context, ~14 % of a full ritual session and ~40 % of a worker's start. It is also the most-synced file in the framework, so a 40 % rewrite lands on every fork at the next `/aios:update` and would conflict with any open PR touching it (#91 does).

The argument is not the tokens. On a live vault, three observed files were condensed (−42 % to −51 %) by folding amendments into the current rule and cutting narrative while keeping every mechanic, path and command; then the old and new versions were run against behavioural scenarios on a frontier model with a pre-registered rubric — **6/6, indistinguishable, at 27 % lower cost per answer**. The same session had four loaded rules fail to fire on the *long* version. **A rule stated in six lines fires; the same rule in forty does not.** `CLAUDE.md` has sections that grew by accretion (Spawning Sessions, Session End, the observed-context rules) and carry the incident that motivated each sentence.

## What the condensation would keep and drop

- Keep: every rule, every command, every path, every section anchor other docs link to, the "why" in one clause.
- Drop: incident narratives that belong in a vault's own antifragile file, restated constants (the file already warns against those), and duplicate justifications.
- Verify: the same scenario battery, old vs new, pre-registered — plus a grep that every backticked literal and every `§` anchor survives.

## Ask

Is a canonical condensation welcome, and on what base — after #91 merges, or as a single PR that absorbs it? If welcome, the author will run the scenario verification before opening it and attach the numbers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019w8zEhSUTUdCn5pRB2v4sE

chuycepeda:
**Yes — welcome, and the argument is the right one.** With one structural condition on how it lands.

[...]

PR #98, opened 2026-09-07
docs(claude-md): condense § Spawning Sessions (46→36 lines), guarded by an integrity test

matiasmacera:
Refs #93 — first section of the section-by-section pass, on the conditions agreed there: base `main` (311b75e), rubric before numbers, anchors and literals as CI, one-clause reasons kept.

## The rubric — written before the condensed text existed

Ten task-shaped scenarios, one per rule the section must make a session *apply*. Two arms, same model (Claude Fable 5.1), same prompt shape: the section — old or new — preloaded as the **only** context, then the operator's question. PASS is read by hand against the criterion below, never by keyword. The condensation is accepted only if the new text passes every scenario the old text passes.

| # | Scenario (prompt, verbatim) | PASS if the answer… | Bonus |
|---|---|---|---|
| S1 | "Necesito un worker que revise los financials del Q1. `~/.aios/spawn-inbox/` existe. ¿Qué hago exactamente?" | writes a request **file** in the inbox (name + task) and does **not** call `spawn` | `tier`/`model` by cognitive load; `surface` derived by pid ancestry, omitted on no match |
| S2 | "El worker `informes` está vivo y quiero pasarle un prompt de tres párrafos. ¿Cómo lo mando y cómo sé si llegó?" | inbox `send` verb; one line, or knows a >~1024-byte prompt is delivered as a **pointer file**; verifies from the **transcript**, not from the file vanishing | never `pgrep` / tab titles — registry is `~/.claude/sessions/*.json` |
| S3 | "Voy a escribir el request con `_tried: []` para que arranque limpio. ¿Está bien?" | **No** — never hand-write a field beginning with `_` | names what `_tried` is |
| S4 | "`~/.aios/spawn-inbox/` no existe. ¿Llamo a `spawn` yo?" | **No** — hands the operator the exact `spawn <name> "<task>"` line (or the Glass button) | names the tell (red-dot / no terminal) |
| S5 | "Spawneo un worker para armar un deck para inversores de una venture. ¿Qué pongo en el brief sobre dónde deja el archivo?" | routes explicitly to `03 - export/{venture}/`, never reflections | `reflections/_inbox/` only as the provisional fallback |
| S6 | "Un barrido de 400 archivos para renombrar wikilinks. ¿Qué tier le pongo al worker?" | `fast` or `scale`; reason: shape of work, not importance | cites the ~22× spread |
| S7 | "Quiero que el worker corra en Gemini para que salga barato. ¿Cómo lo configuro?" | **cannot** — a worker is always Claude; another family is called **as a tool** (`hooks/openrouter.py`); a proxy in the request path is a no | names the reason (third-party code in the path of live credentials) or the judge use-case |
| S8 | "Voy a correr `claude -p` con `--allowedTools Read Grep` para una medición que no debe escribir nada. ¿Alcanza?" | **No** under auto mode — add `--permission-mode default` (or `plan`) | prompt first (variadic flag); only an **absent side effect** is evidence |
| S9 | "Quiero que todos los workers de hoy salgan en Fable. ¿Exporto `CLAUDE_MODEL` en `~/.zshrc`?" | **No** — `spawn --model <id>` per spawn; the global export is the footgun | names why (every future terminal pinned) |
| S10 | "El worker `x` está colgado. ¿Le hago Cmd+W?" | `spawn-kill x`; not SIGTERM, not Cmd+W | names why (respawn loop / orphan to launchd) |

## Verdicts (hand-read)

| # | old section | new section |
|---|---|---|
| S1–S10 | PASS, bonus met | PASS, bonus met |

**Old 10/10 · new 10/10 · bonus 10/10 both.** No scenario where the old text passes and the new one fails, so nothing was dropped. The twenty answers are available on request (they are long; the summary is honest because every one was read).

## Numbers (after the rubric, as agreed)

| | old | new |
|---|---|---|
| section | 46 lines · 14,918 bytes | 36 lines · 10,951 bytes (−27%) |
| context per turn, median of 10 (section + prompt + system) | 28,152 tokens | 26,893 tokens (−1,259) |
| experiment cost | 4.26 USD | 3.91 USD |

Every session loads this section on every turn; the ~1,200 tokens are paid on each.

## What changed in the text

Rules and their one-clause *why* stay — including the ones that are easy to mistake for trivia (`spawn` needs `surface`, `send`/`kill` do not; compare pids, never process names; a vanished request file means picked up, not done; `.undelivered` is a task still owed). What went: the dated incident narratives (diagnosis dates, "hyperfrustrated operators", the 2026-07-23 story), the repeated framing, and the two sub-callouts on `.mcp.json` and OAuth, which are now one paragraph inside `--profile`. One literal was **added**: `surface:pid`, the shape of a `_tried` id, which the old text named only in passing.

## What guards the next section — `tests/claude-md-integrity.test.sh`

Two checks, both greps, wired into `validate.yml`:

1. **Every `§` reference in the repo resolves** to a heading in `CLAUDE.md` — 41 references across 327 files today. A bare roman numeral (`§ VI`) is accepted; a numeral followed by a capitalised section name (`§ V. Vault Commands`) must match that name too, so a renamed sub-section fails rather than hiding behind the numeral.
2. **95 pinned literals are present** — `tests/fixtures/claude-md-literals.txt`, one per line, seeded from the section *before* the edit: every path, flag, JSON shape, command and bold rule. A reviewer diffs the fixture, not the prose.

Green on the untouched `main` text and on this branch. The next section of #93 will extend the fixture with its own literals before condensing.


chuycepeda:
Merging. The rubric-before-numbers discipline is exactly what was asked for on #93, and the integrity test is the part that outlasts this PR — I mutation-tested it rather than watching it pass: deleting a referenced heading is caught and names all five broken references repo-wide, a substantive rename is caught, and deleting a 914-byte prose rule is caught because it takes five pinned literals with it. It then caught **my own** edit to a pinned literal while I was working on top of this branch, which is the review surface doing its job.

Three things were fixed on the branch. **All three were already wrong on `main`** — condensing the section is what put them under a microscope, and none is a correction of the contribution.

**1 · The `hash:` field cited `c5f5df2`, which can never exist on `main`.** The repo squash-merges, so a branch head is unreachable afterwards — and `/aios:update` clones `--single-branch main`, where I measured the object absent while `ba2a010` and `d3b4457` are present. An unreachable hash returns exit 128, which the spec reads as *inconclusive, not synced* — so this entry would have read **NEW forever, for every operator, on every future sync**.

This one is ours, squarely: `CONTRIBUTING.md` said *"every `hash:` must resolve in canonical git,"* and `c5f5df2` **does** resolve — in a full clone. Resolving is a cousin of the real requirement, which is *reachable from `main`*. Fixed in this PR to say so, and to state that the value only exists after merge. A maintainer hit the identical friction on #96/#97 an hour earlier, so the doc was the problem, not the contributor.

**2 · Availability was an existence test where it needed to be a liveness test.** Both this section and `orchestration-ladder` gated on *"if `~/.aios/spawn-inbox/` exists."* That directory persists forever once created, so an operator who quit the App or closed the IDE has the directory and no fulfiller — and the request is never picked up, never dead-lettered, therefore never surfaced (`/today` and `/close-day` report `.undelivered`, not *unclaimed*). Measured on a live machine: one `~/.aios/surfaces/*.json` named a pid dead three weeks, sitting beside another's live one. Now: available iff some `surfaces/*.json` names a running pid — the same artifact already walked for `"surface"` derivation, asked a different question.

**3 · The skill described a world with one surface** — eleven mentions of Glass, zero of the App, including *"no Glass → hand it to the operator."* A session on the App would follow that into asking a human to do what the App fulfils natively. Its rung vocabulary had also gone a version stale (a two-rung `"tier"` choice after the ladder grew to four); that vocabulary is now deleted rather than corrected, so `MODEL-ROUTING.md` owns rungs and ids alone and the skill cannot fall behind a list it does not restate.

The section lands at **−16 %** rather than −27 % — about 1.5 KB went back into the liveness rule, the restored *"un-sandboxing does not help, the classifier gates it"* clause, and naming the inbox `kill` verb before `spawn-kill`. Deliberate trade: this wiring decides how every AIOS session delegates work, so shorter is only worth having if it is also right.

**One limit now stated in the test header**, so nobody over-reads a green build: the manifest pins literals, never the one-clause *why*. I removed 279 bytes of pure reason-clause with zero literals touched and both checks stayed green. The rubric is what guards the reasoning and it needs credentials, so it cannot run in CI — green means *no literal was dropped*, not *nothing was lost*. Please keep running the rubric per section, as you planned.

For the next section: base on `main` at the time you start, cite the PR number in the field, and leave the hash to the merge commit — I'll add it in the follow-up, same as here.

PR #111, opened 2026-09-09
fix(quota-watch): adoption check uses the triggering metric and notifies once per swap

fernandezdiegoh:
## What

`check_adoption()` in `hooks/claude-identity/_watch.py`:

- compares the metric that **triggered** the rotation (`7d …` reason → `seven_day_pct`, otherwise `five_hour_pct`) instead of always the 5h window;
- notifies the desktop **once per rotation**: the alert file is keyed by the swap's `ts`, and a tick that finds the same `ts` already recorded logs but does not re-notify.

## Why

A rotation fired by the 7d cap lands on an account whose 5h baseline is unrelated to the source's, so "5h did not fall by 3 points" is not evidence of a failed adoption. And the fast path kicks the watcher every 30 s (3 s in the hot zone) with no dedupe, so a (false) failure became a notification storm.

Measured 2026-09-08 23:30–23:37 on a 3-account setup: gmail → sovra on `7d at 98%`; gmail 5h 7 %, sovra's own 5h 21–35 % while the operator worked; 7d went 98 % → 17 % (adopted). **Eleven "adoption failed" desktop alerts in six minutes**, one per tick, on a rotation that had worked.

## Verified

- `py_compile` clean; one real `claude-identity.sh watch` tick after the patch logs `cooldown active … skip` with no `ADOPTION FAILED`, and the stale `rotation-alert.json` is gone (7d 98 → 17 counts as adopted).
- The log keeps one line per tick, as before; only the `notify()` is deduped. The alert file gains `swap_ts` and `metric`; the statusLine renderer only reads `kind`/`message`, unchanged.

Second commit: an operator **pause marker** (`CONFIG_DIR/quota-watch.paused` holding a future epoch or ISO-8601) that suspends rotation and the adoption check until it expires; the marker removes itself once past due. There was no kill switch before: the statusLine fast path has no operator-controlled env, and unloading launchd covers only one of the two entry points.

Sibling of #83 (same operator setup).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,14 @@ The probe now reports **entries, highest number, and duplicates** in one pass, c
 
 **Action required:** none. Next time `/aios:compact` runs it measures this way. If it reports a mismatch, stop and diff before letting it compact.
 
+### Eleven "adoption failed" alerts in six minutes, for a rotation that worked
+
+**What you can now do.** Stop getting that. When the quota autopilot rotates accounts, it checks afterwards that the live session actually picked up the new credential — and it was checking the wrong number. A swap triggered by the **7-day** cap was judged on whether **5-hour** usage fell, and those two have nothing to do with each other: the new account's 5h baseline reflects whatever you were doing on it, not the account you left. So a rotation that worked perfectly reported failure, once per tick.
+
+It now compares the metric that actually triggered the swap. And a failed adoption is a **state, not an event** — the log still records every tick, but your desktop gets **one** notification per rotation instead of one every three seconds.
+
+**Action required:** none.
+
 ### An entry could ship with an empty `hash:` and read as new forever
 
 **What you can now do.** Nothing to change — this one is about entries reaching you correctly. `CONTRIBUTING.md` told contributors to cite the PR number and add the hash after merge, without saying what the field holds meanwhile. It now says: **leave the line present with an empty value, never omit the line** — because the two mistakes are not symmetric. An empty value keeps the entry *shown* until a maintainer fills it. A missing line has nothing to test, so "already synced" is vacuously true and the entry is **silently skipped for every operator, forever.**
--- a/hooks/claude-identity/_watch.py
+++ b/hooks/claude-identity/_watch.py
@@ -219,8 +219,19 @@ def check_adoption(cache: dict) -> None:
     # would report a failure that has not had a chance to happen yet.
     if (cache.get("captured_at") or 0) <= (last.get("ts") or 0):
         return
-    before = last.get("five_hour_pct") or 0
-    current = cache.get("five_hour_pct") or 0
+    # Compare the metric that TRIGGERED the rotation, not always the 5h window.
+    # A swap fired by the 7d cap lands on an account whose 5h baseline has nothing
+    # to do with the source's, so "5h did not fall" is not evidence of anything.
+    # Measured 2026-09-08: gmail rotated to sovra on 7d 98%; gmail's 5h was 7%,
+    # sovra's own 5h sat at 21-35% because the operator was using it — eleven
+    # "adoption failed" alerts in six minutes while the 7d had gone 98% -> 17%.
+    reason = last.get("reason") or ""
+    if reason.startswith("7d"):
+        key, label = "seven_day_pct", "7d"
+    else:
+        key, label = "five_hour_pct", "5h"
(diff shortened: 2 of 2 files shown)


chuycepeda (changes requested):
The headline fix is right, well-evidenced, and I verified it end to end. The PR just carries more than its title, and that third thing needs its own door.

## What I verified and want to keep

**The metric fix is correct.** `reason` is written as `f"7d at {pct7}% (>= {t7}%)"` (`_watch.py:422`) and `f"5h at …"` (`:420`), so `reason.startswith("7d")` fires as intended — including the composed form at `:440` (`f"{reason} | {why}"`), which still starts with the metric. And `seven_day_pct` genuinely exists on both sides of the comparison (`:314`/`:316` write it into the rotation record, `:441`-`:463` into the cache), so `before`/`current` are real numbers rather than a pair of zeros. That last one was the check I most wanted to fail, because if the key were absent the comparison would silently report "adoption failed" on *every* 7d rotation — the inverse of the fix.

Your diagnosis is exactly right: a 7d-triggered swap lands on an account whose 5h baseline has nothing to do with the source's, so "5h did not fall" was never evidence of anything.

**The notify-once change is correct too**, and the reasoning in the comment is the part I'd have asked for — *a failed adoption is a STATE, not an event*. Keying the dedupe on `swap_ts` also degrades correctly: an alert file written before this change has no `swap_ts`, so `.get("swap_ts")` returns `None`, the comparison fails, and the operator gets exactly one notification after upgrading. That's the right migration behaviour and I don't think it was an accident.

## What has to come out

The PR also adds **`PAUSE_FILE` / `paused_until()`** — a new operator-facing capability that suspends rotation. That's not a fix, it isn't in the title, and as shipped an operator cannot discover it: no documentation anywhere, no test, no CHANGELOG entry, and no stated format (the code accepts an epoch float *or* ISO-8601, which a reader has to infer from the source). Canonical's rule here is the reviewer half of `custom/`-first: **an undocumented feature in shared code is a personalization**, however useful.

I want the feature — the auto-expiry design is genuinely thoughtful, and *"a forgotten pause cannot silently disable the autopilot forever"* is the right instinct. It just needs its own PR with the format documented where an operator will find it, a test, and a CHANGELOG entry saying how to use it.

## Also missing

No CHANGELOG entry, and this changes operator-visible behaviour (they stop getting eleven notifications). One consolidated entry per day — today's is `2026-09-09`, so add a subsection there rather than a new heading.

**So: split it.** Drop the pause block from this PR, add the CHANGELOG subsection, and I'll merge it today. Open the pause as its own PR and I'll review that today too.

chuycepeda (approved):
Ignore my earlier request — I did the split myself rather than leave you holding work we already knew how to finish. Your fix is unchanged and still yours; here's exactly what moved so nothing is a surprise:

**On this branch now:** your `2a7881d` alone, rebased onto current `main`, plus one commit from me adding the CHANGELOG entry it was missing. Your authorship is intact on the fix.

**Moved to #118:** your pause commit, cherry-picked with your authorship preserved, plus the documentation and test it needed — README § *Pausing the autopilot temporarily* (both formats, macOS and Linux commands, and the note that the path honours `CLAUDE_CONFIG_DIR`), a row in § Files written, and `tests/quota-watch-pause.test.sh`. Nothing of yours was dropped.

I force-pushed this branch to do it, which I'd normally ask before doing — `maintainerCanModify` was on, the pause commit was already safe in #118, and I fetched your fork first so the lease had a real basis rather than reaching for a bare `--force`. Say the word if you'd rather own that step yourself next time.

**On the substance — the pause design is the part I'd single out.** *"A forgotten pause cannot silently disable the autopilot forever"* is the right instinct, and your implementation already got every case right on the first run: future epoch and future ISO-8601 pause and persist; past, malformed, empty and whitespace-only markers are ignored **and removed**. I only pinned that behaviour in tests, I didn't have to change it. That asymmetry is the whole safety property, because a pause is the one operator capability whose failure mode is silence.

And the diagnosis on the metric was exact: a 7d-triggered swap lands on an account whose 5h baseline has nothing to do with the source's, so *"5h did not fall"* was never evidence of anything. I checked the detail that would have inverted the fix — `seven_day_pct` exists on both sides (`:314`/`:316` into the rotation record, `:441`–`:463` into the cache) — so it's comparing real numbers, not two zeros. Approving on that basis.

Thanks for both of these. #118 is up for review whenever you want to look.

PR #112, opened 2026-09-09
fix(compact): Step 3.5 entry probe measures structure, not a bare count

chuycepeda:
## What

`/aios:compact` Step 3.5a prescribed `grep -c '^### [0-9]'` as the antifragile entry count. That probe fails two ways:

1. **Over-count** — it counts numbered meta-pattern index headings (`### 1. The rush-to-commit pattern`) and date-prefixed headings as entries (~2× on vaults with a numbered index).
2. **Concealment** — on a file carrying a complete stale copy of itself, it reads as a big-but-plausible number instead of *damaged*. A real vault measured **207 against 107 distinct** while doubled; the defect survived 11 days and 8 commits, and a compaction pass run against it would have tombstoned into the copy carrying the newest content.

## The fix

- Step 3.5a now ships a **structural probe** (fenced, extractable): entries counted **outside** the `## Meta-patterns` section, reporting `entries / highest / duplicated` in one pass, with the invariant **entries == highest AND duplicated == 0** and a hard stop-and-diff on mismatch.
- The index-size boundary is now the **section** (→ next `## ` heading), never a heading shape — the index may be numbered or lettered.
- `tests/compact-entry-count.test.sh` — 9 checks: the prose carries the invariant; the probe is extractable and section-scoped; the **document's own probe** (extract-never-restate) runs against incident-shaped fixtures (numbered index + glued stale copy); a control asserts the retired bare count still misreads them, so the fixtures can't silently stop reproducing the incident.

Closes the deterministic-catch half that AI-139's rule requires for this finding.


chuycepeda:
**For the merging session (aios-canonical):** this PR deliberately ships **no CHANGELOG rider** — the subsection pushed 2026-09-08's entry past the 1,500-word budget (its own guard caught it), and the entry belongs to your merge day anyway. On merge, add one short subsection to your day's entry, e.g.: *`/aios:compact` — Step 3.5's probe now reports `entries / highest / duplicated` and stops on mismatch; a file doubled by a bad edit can no longer read as a plausible count. Nothing for operators to do.* Then append the merge hash per the any-hash rule.

PR #115, opened 2026-09-09
fix(close-day): launch the two stdlib hooks with uv run, like /today already does

s1sko7:
`/today` invokes the bus dead-letter check with `uv run`; `close-day.md` still invokes the same script with `python3`. On Windows `python3` resolves to the Microsoft Store alias, which prints an install notice and exits 49 — never a result. So the check that exists to end silent failures cannot fire at all on Windows, and the sibling command already had the fix.

**Measured** on Windows 11 (PowerShell 7, uv 0.11.19):

```
$ python3 --version >/dev/null 2>&1; echo $?
49
$ uv run ~/aios/hooks/bus-dead-letters.py
bus-dead-letters: none        (rc 0)
```

Two invocations change, and only two:

- `bus-dead-letters.py` — the exact line `/today` already ships with `uv run`.
- `buffer-status.py` — same file, same failure, missed when the first was fixed.

### Why the scope stops there

`uv run` is only interchangeable with `python3` for a script that needs no third-party packages, because **without PEP-723 metadata uv resolves no dependencies**. Both scripts above import stdlib only, so the swap is safe.

`ingest.md` also invokes `python3` three times and is broken on Windows the same way, but it is **not** included here: `markitdown-convert.py` imports `markitdown`, so `uv run` would trade a Windows-only failure for one on every platform. That one needs PEP-723 headers on the hooks, which is a separate change — happy to open it if useful.

No new dependency: `uv` is already required by the pipeline-executor step that runs earlier in the same command.

https://claude.ai/code/session_01Xj6ZYu64e5sfHKZibPgkMQ


chuycepeda (changes requested):
Thanks for this — and the instinct is a good one: one launcher convention across the commands is worth having. But I can't take it as written, for three reasons that are checkable rather than stylistic.

**1. The premise doesn't hold.** The title says *"like `/today` already does"*, but `/today` invokes this same script with `python3`:

```
plugins/aios/commands/today.md:54
- `Bash(python3 ~/aios/hooks/bus-dead-letters.py)` — **bus dead-letter check.** …
```

`/today` uses `uv run` only for `pipeline-executor.py`, and that's for a concrete reason — it's the one script with a PEP-723 block:

```
hooks/pipeline-executor.py:2   # /// script
hooks/pipeline-executor.py:4   # dependencies = [
```

`bus-dead-letters.py` and `buffer-status.py` are plain stdlib (`#!/usr/bin/env python3`, no script block, no dependencies), so there's nothing for `uv` to resolve.

**2. It creates the drift the edited line exists to prevent.** That line reads *"same script `/today` runs — one implementation, so the two surfaces can't drift."* After this change the two commands invoke the same script two different ways, which is the drift, just moved from the script to the launcher.

**3. It leaves a contradiction inside the line it edits.** The same sentence still says *"requires `python3`"*.

For what it's worth I ran both, and neither is broken — `python3 hooks/bus-dead-letters.py` and `uv run hooks/bus-dead-letters.py` both print `bus-dead-letters: none` and exit 0. So this isn't a bug fix; it's a convention change presented as one.

**If you want the convention, I'll take it — as a complete change:** both commands (`/today` and `/close-day`), the `requires python3` prose updated wherever it appears, and one sentence in the PR saying *why* `uv run` is preferable for a stdlib script. The portability argument is the strongest one available (a machine with `uv` but no `python3` on PATH) — if that's your reasoning, say so and it becomes a real fix with a real justification.

Either widen it or close it, whichever you prefer. Happy to review the wider version today.

chuycepeda:
Following up to sharpen one thing I said, because I was too flexible and that would have cost the design.

I closed with *"widen it and I'll take it."* I'm withdrawing that. Here's the measurement that changed my mind:

```
PEP-723 hooks (declare dependencies → uv run is REQUIRED):  1   (hooks/pipeline-executor.py)
stdlib-only hooks (python3 is correct):                     20
```

This PR converts **2 of those 20**. So it doesn't achieve consistency even within its own class — it creates a *third* state where two stdlib hooks launch one way and eighteen launch another. Widening it to all 20 wouldn't fix that; it would just move the arbitrariness, and buy a hard `uv` dependency on twenty paths that have nothing for `uv` to resolve. There's also no `pyproject.toml` at `~/aios`, so `uv run` on a script with no dependency block resolves ad-hoc rather than against anything meaningful.

**The rule the codebase already follows is better than a house style, and it's the one I want to hold: the launcher is a property of the script, not a convention.** A script declares its dependencies, and the launcher follows from that declaration — `uv run` if there's a `# /// script` block, `python3` if there isn't. That rule is *derivable from the file in front of you*, which means the next hook anyone adds answers its own launch question with no doc to consult and nothing to remember. A house style has to be looked up, and drifts the moment someone doesn't.

So my posture before you decide: **I'm not taking a launcher change on these two, and I'm not taking the widened version either.** That isn't inflexibility about your PR — it's that "one launcher for everything" is a worse rule than the one we have, and I'd be trading a self-documenting property for a convention.

**What I would take, gladly, is the problem underneath it if there is one.** The only argument that beats the current rule is portability: a machine with `uv` installed and no `python3` on `PATH`. If that's what you actually hit, say so and it becomes a real finding — but then the fix isn't editing call sites, it's a launcher shim or a documented prerequisite, and it's a much more interesting PR than this one. If you didn't hit it and this was a consistency instinct, that's a completely reasonable instinct and the answer is just that the consistency already exists one level down.

Your call whether to close it or come back with the portability case. Either is a good outcome and no hard feelings on a two-line PR that made us write the rule down.

chuycepeda:
Closing this one for now — no hard feelings and genuinely no judgement on the instinct behind it.

To recap where it landed, so reopening is easy: the measurement that decided it was that **1 hook declares dependencies** (`hooks/pipeline-executor.py`, PEP-723) and **20 do not**. This PR converts 2 of those 20, so it doesn't reach consistency within its own class — and widening it to all 20 would buy a `uv` dependency on twenty paths that have nothing for `uv` to resolve. The rule already in the codebase is the better one because it's *derivable from the file in front of you*: a script declares its dependencies, and the launcher follows from that declaration.

**The one argument that would beat it is still open**, and I meant the offer: if you actually hit a machine with `uv` installed and no `python3` on `PATH`, that's a real portability finding — and the fix then isn't editing call sites, it's a launcher shim or a documented prerequisite. That's a more interesting PR than this one and I'd review it the same day.

**Reopen this any time**, or open the portability version, whichever fits. Nothing here is settled against you — a two-line PR that made us write the rule down did more than most.

PR #122, opened 2026-09-11
fix(freshness): report a refused login as access-denied, not "offline"

Pipena:
## What

`/today`'s framework and company freshness checks, and `/close-day`'s framework check, ran `git ls-remote … 2>/dev/null` and mapped any empty result to `unreachable`, which both commands render as *"offline — fine for now"*. Throwing away stderr made a refused credential indistinguishable from an outage.

This moves the checks into `hooks/freshness-probe.sh`, which keeps git's error output and adds one state, **`access-denied (why)`**, when the server answered and refused the credential (`Permission denied` · `Repository not found` · `could not read Username` · `Authentication failed`). Everything else stays `unreachable`. The rendering rules in `today.md` and `close-day.md` now surface `access-denied` as a task; only `unreachable` may be called offline.

## Why

Observed on a live install: a company-context check read *"unreachable — offline, not a problem"* on most mornings for about a month, while the network was fine and GitHub was refusing the machine's SSH key (the account had also lost access over HTTPS). Nothing in the output could have told the operator.

## Test

`tests/freshness-probe.test.sh`: 15 checks, no network needed (a stub `git` on `PATH` emits the real error strings). It includes a **control** that runs the old inline probe against the same stub and asserts it prints `unreachable` for the refused repo, so the suite demonstrably separates the two behaviours. Passes under GNU bash 5 and macOS bash 3.2. `tests/changelog-entry-shape.test.sh` passes with the new entry.

## Notes

- Output words are unchanged for the existing states, so anything that reads `synced` / `BEHIND` / `unreachable` / `no-config` keeps working.
- The SSH → HTTPS fallback and the `find`-not-glob company scan are kept as they were.
- CHANGELOG entry added with `hash: ` left empty for the maintainer, per CONTRIBUTING.


chuycepeda (approved):
Approving, with one commit added on top rather than handed back — the change is small enough that asking you to round-trip for it would cost more than it's worth. Everything below is what I checked, since you measured your claims and that deserves the same back.

## Verified

**The premise holds on `main`.** The inline probe is `git ls-remote … 2>/dev/null` with empty → `unreachable`, and `today.md:71` renders that as *"offline — fine for now."* A refused key and a dead network are genuinely indistinguishable there. Your month of false "offline" mornings is exactly what that code does.

**The thing I went looking for is handled.** My first concern was ordering: if SSH is refused but the HTTPS fallback works, does it wrongly report `access-denied`? It doesn't — `probe()` returns `synced`/`BEHIND` before classifying, and only reaches `classify` when both attempts produced nothing. That's the right shape and it wasn't obvious.

**`GIT_TERMINAL_PROMPT=0` + `BatchMode=yes` + `ConnectTimeout` is an improvement beyond the stated scope.** The inline version could sit waiting on a credential prompt; this can't. Worth knowing you fixed something you didn't claim.

**15/15 under bash 5 and under bash 3.2** (macOS system bash), confirmed on both here.

**And the control is the part I'd have asked for if it were missing.** Running the *old* inline probe against the same stub and asserting it prints `unreachable` is what makes the suite evidence rather than assertion — it demonstrates the two behaviours are actually separated, instead of only showing the new one works. Same for `find`-not-glob and the `</dev/null` on the inner `probe` call inside the `while` loop; both are the kind of detail that only shows up when someone has actually been bitten.

## What I added

One defect, narrow and real. `reason()` took the first match of `ACCESS_RE|NET_RE` while `classify()` read the whole accumulated stderr. Both attempts' errors are concatenated, so on a network with **port 22 blocked and 443 open** — ordinary in a corporate environment — the SSH error is a network one and the HTTPS refusal is what decides the classification. Output:

```
access-denied (Connection refused)
```

The classification is right; the quoted reason came from the other attempt. A line that contradicts itself sends the operator to check their wifi over a credential problem — the same confusion this PR exists to remove, one layer in.

`reason()` now takes the classification and searches only that class's pattern. Three checks added, and they **extract** your shipped `reason()` and its regexes rather than restating them, so they can't go green while the code drifts. Control verified: reverting to the first-match version fails the suite. **18 passed on bash 5 and 3.2.**

## Note

You left `hash:` empty per `CONTRIBUTING.md`, which is correct and is the convention that only got written down two days ago (#117) — partly because earlier PRs kept hitting the gap where it wasn't. Good catch reading it.

Merging now. Thank you for the month of mornings you spent noticing this rather than filtering it out.

Pipena:
Thanks Chuy, for the merge and for the review itself. The reason() catch is exactly the kind I would have missed: the right verdict, with the wrong evidence quoted. I've ported it to my local copy, with a test for the port-22-blocked case.

Luigi (drafted with ALI)

PR #125, opened 2026-09-11
AI-150 Half A: mirror the resume verb into canonical's docs (unblocked — App v0.9.5 shipped)

chuycepeda:
> ## ⛔ Hold this PR
>
> **Do not merge until the App has shipped `AI-149`** (currently 🔴). Canonical only *documents* the bus verbs — the App implements them. Landing this first ships a documented verb with no fulfiller behind it, which is the exact class `CONTRIBUTING.md`'s symptom router exists for and the one `AI-129`/#100 already cost us once.
>
> Opened now rather than held as a local branch so CI has already judged it and merging is one click the day the App lands.

## What this mirrors

The App built `resume` on 2026-09-11 and recorded the shipped semantics. This mirrors **those**, not the sketch `AI-150` was minted with — the sketch predates the build and was missing two things canonical has to state.

- **`{"action":"resume","name":"<kebab>","prompt":"<first prompt>"}`** reopens a **closed** named session.
- **The name resolves from the transcripts** (`~/.claude/projects/*/*.jsonl`), never the session registry. A renamed session resolves by its **last** `agent-name` record — the same record `CLAUDE.md`'s Mandatory First Action already greps, read last rather than first, so this is one mechanism described once.
- **No transcript → dead letter naming the miss**, never a silent spawn. `"fallback":"spawn"` is opt-in and is a **string**, not a boolean, so a second fallback later needs no second field.
- **Already running → revealed and the prompt delivered**, exactly as `send` would; resuming a live session would give one identity two processes.
- **An unrecognised `action` is now REFUSED.** An absent `action` still means spawn (the `{name, task}` form is untouched). This one is in the **prose**, not just the verb list, because it changes what every agent `CLAUDE.md` instructs to write these files must do: before `resume`, degrading an unknown verb to spawn was harmless — spawn was the only thing it could have meant. Now the outcomes differ, so a typo'd `"resmue"` would have handed back a brand-new session in place of the one you asked to reopen, context gone, nothing reporting the substitution.
- **`INBOX_CONTRACT` 2 → 3** in both documents. A fulfiller still on 2 handed a `resume` spawns a duplicate, so the stamp is how a reading agent tells which rules the local fulfiller follows.

## A contradiction the mirror exposed

`CLAUDE.md` said, unqualified: **"Addressing — the session registry is the only truth."** `resume` falsifies that. The registry lists only *running* sessions, so a closed one is absent from it **by design, not by omission** — and an unqualified rule sends an agent to the single source that structurally cannot answer the question `resume` exists to ask. Now scoped to live sessions, with why `resume` differs and an explicit note that "absent from the registry" never means "no such session existed".

Mirroring a new verb without catching this would have left canonical stating a rule its own new verb breaks.

## Not touched, deliberately

`~/.aios/spawn-inbox/README.md`. `CLAUDE.md` describes it as *rewritten by whichever surface boots so it cannot drift from the running code* — generated, not authored. Canonical shipping one would manufacture exactly the drift the generation prevents. The suite asserts canonical contains no such file.

## Tests — `tests/bus-verb-parity.test.sh` (13)

The check `AI-150` asked for. The verbs live in two hand-maintained copies — `CLAUDE.md` (loaded by every session at startup) and the `orchestration-ladder` skill (loaded when a session is deciding how to delegate) — and nothing compared them. The failure mode is worse than drift: the two are read at **different moments**, so a verb in the skill and not in `CLAUDE.md` is invisible to a session that never had reason to load the skill.

It asserts **agreement**, plus the presence of the behaviour-changing rules and that the contract *numbers* match — never a canonical-side verb list, which would be the third copy of the thing being prevented. Controls: an extractor that finds fewer than 3 verbs fails as broken-extractor rather than passing vacuously, and a mutated fixture with a verb added to one list only must be detected.

49/49 suites green.

## Merge checklist (for whoever lands this)

- [ ] `AI-149` is shipped in the App and the verb answers on this machine
- [ ] Re-read the App's final `AI-150` notes — if the shipped shape moved after 09-11, fix the mirror before merging
- [ ] **Add the `CHANGELOG.md` entry, dated the actual merge day** — deliberately not in this PR, because a dated entry written now would sort wrong and claim a day this did not ship
- [ ] Flip `AI-150` Half A on the roadmap; leave Half B's status alone (it shipped separately in #124)


chuycepeda:
**Unblocked.** App **v0.9.5** shipped `2026-09-11 22:26 UTC` carrying `AI-149`.

Merge-checklist results:

- [x] **`AI-149` shipped and the verb exists** — verified against the shipped tree on `origin/main`, *not* against this machine. The operator is deliberately holding the update to screenshot the pending-update state, so the local `~/.aios/spawn-inbox/README.md` is written by the running (older) handler and is not evidence of what shipped. `src/core/busVerbs.ts` → `BUS_VERBS = ['spawn','kill','send','resume']`; `src/core/sendQueue.ts` → `INBOX_CONTRACT = 3`. The mirrored verb set was diffed against that array: **identical**.
- [x] **Re-read the App's final notes — the shape had moved, and two things in the mirror were wrong:**
  - The stamp an agent can actually *find* is an HTML comment reading **`aios-spawn-inbox: contract N`**. `INBOX_CONTRACT` is the symbol's name inside the App/Glass source and appears **nowhere** in the README — so telling an agent to check `INBOX_CONTRACT` described a check that cannot be performed. Same class as the registry-only-truth contradiction this PR already fixes.
  - `normalizeVerb()` trims and lowercases before matching, so `"Resume"` and `" resume "` resolve. Left unstated, an agent reading a refusal blames its own formatting instead of concluding the verb is unimplemented on that surface.
- [x] **Changelog entry added, dated the actual merge day** (today — so it folds into the existing `2026-09-11` entry rather than minting a second heading).
- [ ] Flip `AI-150` Half A on the roadmap — doing that at merge.

The parity suite grew 13 → **15**: both docs must now name the stamp *as it appears in the file*, and both must state the case/whitespace rule. The rebase kept **both** this branch's CI step and #126's — they guard different things, so the union was the correct resolution rather than either side.

50/50 suites green, bash 3.2 verified.

PR #127, opened 2026-09-12
Instruments that mislead: note-append anchoring, canonical hygiene, voice-gate v2

chuycepeda:
Reported as *"today's note is a complete mess"* — three sections in the wrong place and one apparently **emptied**. Diagnosed from the git history and the session transcript rather than inferred.

## What happened

`/close-session` inserts with `aios-note-append --before "## Close of Day"`. The helper matched with a bare `index($0, marker)` — **anywhere in a line**.

A daily note is prose, and prose quotes headings. A bullet written that morning contained a sentence explaining this helper's own behaviour, quoting the heading inside it. Hours later a close-session ran `--before` against a note that had **no such heading** — which should have appended at the end. Instead it anchored to that bullet: the block landed mid-note and split the section it landed in, heading above the block and its two bullets stranded below.

**Nothing was deleted and nothing errored.** Verified line by line: 71 lines of content before, 71 after, zero lost. The "emptied" section was its body pushed below an inserted block. That invisibility is the whole problem — the note just read wrong.

This is **antifragile #105's family, one domain over**: *"the moment a fix is worth documenting, the documentation lands in the same file the guard scans."* There the instrument was a grep scanning source; here it is an insertion anchor scanning a note. Same trigger, and the same reason writing it more carefully does not help.

## The fix

- **The marker must match at position 1.** A heading marker starts with `#`, and a prose line starting with `#` *is* a heading, so anchoring is exactly right for this marker class.
- **One matcher decides both branches.** The retired version asked `grep -qF` whether to enter the `awk` at all — so two different matchers had to agree about a single question, and a marker one found but the other would place differently chose the wrong branch silently. The no-match case is now the `awk`'s own `END`.

## Tests — `tests/note-append-marker.test.sh` (11)

**Refusal cases are written first**, per #110: the bug lived entirely in the should-not-match branch while the happy path was fine throughout.

1. the exact incident — marker quoted inside prose, no such heading → must append at the **end**, section unsplit
2. marker absent entirely → end
3. a heading that merely shares a prefix must not steal the anchor
4. *then* the happy path
5. **control** — replays the retired predicate against the incident fixture and requires it to still mis-anchor
6. structural — no second matcher gating the insert

Assertions compare **heading order, not line numbers**: the first draft asserted line numbers and failed twice on correct behaviour, which is #105's "assert the rule, not the wording" in miniature.

## Also in this PR: de-personalization of six shipping files

Canonical is authored from inside a live vault, so vault-personal references arrive as **good writing** — concrete evidence is what makes a rule persuasive. The audit found dated incidents from one operator's week, a possessive reference to their vault, and example slugs naming real outside parties who never agreed to appear in a public repo. Every rule is kept; only the evidence is generalised.

`tests/canonical-hygiene.test.sh` (8) guards the class **without naming anyone** — hardcoding the offenders would republish them on every run and go stale for any other contributor. It matches the shape, and its allow-list of generic possessors is what keeps `the operator's vault` (correct, everywhere) from being flagged.

**A second check was written and deleted.** It flagged any wikilink to a date-prefixed note — and fired on the corrected example that replaced the real names, because a dated wikilink is how the convention is *documented* and no regex separates an invented surname from a real one. Per #105, a check that cannot tell an example from an instance trains the reader to ignore it. The reasoning is recorded in the file so it is not re-attempted.

52/52 suites green, bash 3.2 verified.


chuycepeda:
**Added `AI-151` — voice-gate v2.** Folding it here rather than opening its own PR: the row said *"fold into #64 if unmerged, else its own PR"* and #64 is merged — but these four changes are **one theme**, which is the better reason. Every one of them is an *instrument that misled*:

- a **marker** that matched prose and turned a sentence into an insertion anchor
- **evidence** that persuaded readers and should never have shipped
- a **guard** that fired on its own fix (written, then deleted)
- a **pattern pass** mistaken for a verdict

### The three gaps

**Per-surface baselines.** Step 1 built *one* baseline from published work — but short-form compresses a long piece to its quotable lines, so it runs several times long-form density *by construction*. A blended number is wrong in both directions at once: false-alarms on every rendering, false-cleans every essay. And the incident class the skill exists for — a published **short post** called AI-written — lands exactly where a blended baseline is silent. Burst granularity follows: **a post is a paragraph**, so the long-form burst rule must not transfer to a surface where every unit is already burst-sized.

**Re-derivation cadence.** Computed once and never revisited, the *corpus* drifts while each draft reads in range — every piece measured against a number the earlier pieces already moved. Now every ~5 publishes or quarterly, with the derivation date recorded beside the number so its age is visible rather than assumed. Same doctrine `CLAUDE.md` applies to derived observed files.

**Candidate-finder honesty.** A mechanical pass finds candidates; it does not return the verdict. Pivots land on subjects a pattern wasn't written for, punctuation hides inside quotes and code — a conservative pass can score **zero** on a page a careful read finds several tells in.

### Hygiene

No operator-specific figures. The skill's whole model is a per-writer budget, and a measured corpus belongs to whoever measured it — so the rules are written from the **format property**, not from anyone's numbers. `tests/canonical-hygiene.test.sh` (added in this same PR) passes over the new text, which is a fair first exercise of the guard.

52/52 suites green, 14/14 CI.

PR #128, opened 2026-09-12
feat(spawn): spawned workers load operator context on demand, with a floor

matiasmacera:
## What I observed

A spawned worker boots with **no operator context at all**. `CLAUDE.md` § Identity & Greeting step 4 says *"proceed with the full Session Start Ritual"*, and in practice that is often executed as none of it — so a worker can answer a question about an operator's ventures having read nothing they ever wrote. The instruction is not wrong; it is just not one a worker reliably follows, and nothing reports when it doesn't.

## What changes

**Step 4 states a rule a worker can follow.** List the two context folders, read their `_index.md`, then read **only the files the task touches** — as the session's first tool calls, not described and skipped.

**The floor is the half that is not an optimisation.** `growth.md` and `patterns.md` are read **every time, regardless of the task**, because they hold what an operator tends to avoid or repeat and *no task ever names them*. Without the floor, "read what the task touches" hands a fresh worker the relevance call, which is the one judgment it is worst placed to make: it cannot know that the file the task never mentions is the file that makes an answer *this operator's* rather than merely correct. That failure is silent — the worker still answers, and still sounds right.

**Two files, because a worker is born on two paths.** `spawn` in a terminal gets the rule injected at the top of its task file by the wrapper. A worker born from a surface fulfilling an inbox request never reads that file and is governed by `CLAUDE.md` alone — so `CLAUDE.md` is the *only* cover for that path, not a duplicate of the wrapper.

## The mechanism

Two hand-maintained copies of one rule is exactly how a floor goes missing from one side. `tests/spawned-worker-context-parity.test.sh` asserts they agree, and is built to avoid two traps it could easily fall into:

- it **derives** the floor from the wrapper instead of keeping its own list — a canonical-side list would be a third copy, which is the thing being prevented;
- it **refuses to pass on an empty floor**, because "every file in an empty set is present" is true and useless.

Both failure directions were verified to turn the suite red before committing: dropping the floor from `CLAUDE.md` fails, and defining the preamble without wiring it into the task file fails. Restored, 7/7 pass.

## Evidence and its limits

Measured on a live vault: **three of three tasks** reached the same or better answers on demand as with a full preload, at roughly **a third of the cost** — a preload writes every context file to the cache on the first request.

Stating the limits plainly: n=3 is thin for a framework-wide change, which is why the scope is deliberately narrow. The **interactive primary session is untouched** and keeps the full ritual — the case for preloading is cross-file synthesis that no question triggers, and that case is untested. Only spawned workers change.

This has been running in production on one vault since 2026-09-07.

## Checks

- `spawned-worker-context-parity` 7/7 (new)
- `claude-md-integrity` 2/2 · `changelog-entry-shape` 17/17 · `install-wrappers` 26/26 · `canonical-hygiene` 8/8 · `bus-verb-parity` 15/15
- `bash -n` clean on the wrapper installer

## Notes for the maintainer

- The `CHANGELOG.md` entry is its own dated block rather than a subsection of 2026-09-11: that day's entry measured **1497 words against the 1500 budget**, so there was no room. Re-date it at merge if you prefer.
- `` `hash: ` `` is left present and empty per CONTRIBUTING.

https://claude.ai/code/session_01723SEBYDhfXWykLok8xq5g


chuycepeda:
Reviewed and **built on top of this branch** rather than around it — your commit is untouched and the PR stays yours.

## What we kept

**The floor is the real fix and it is unchanged.** `growth.md` and `patterns.md` read every time, whatever the task, because no task ever names them. "The worker read nothing" was the actual failure and this ends it. The diagnosis and the evidence-gathering here were both good.

## What we changed — the axis

Keying the rule on **identity** (primary = full ritual, worker = on demand) creates a permanent class of thinner agents. The bundled roster is mostly the wrong class for that: of 42 agents, the communication bundle alone is `content-writer`, `email-drafter`, `deck-builder`, `meeting-prepper`, `content-scheduler` — every one produces output **in the operator's voice, for an audience**. A `content-writer` that skips `personal_voice.md` because the task did not name it writes competent prose that is not theirs, and nothing reports it. That is precisely the failure the `voice-gate` skill exists to catch, so the rule would have manufactured the problem another part of the framework was shipped to detect.

Your evidence cannot settle it either way, and you said as much: three tasks, with no characterisation of what kind. That supports *"on demand works for some tasks"* — not *"on demand is the rule for all worker tasks."*

So the discriminator is now one question about the worker's **own output**, which also answers your own objection that a fresh worker cannot make the relevance call. It cannot rank fourteen files; it can answer one yes/no about its deliverable:

> *Will what I produce be read as the operator's own words, or act on their behalf?*
> **yes** → full ritual · **no** → floor + what the task touches · **unsure is not a third answer, read everything**

The asymmetry is the argument: over-reading costs one session's tokens, under-reading costs the operator's voice and fails silently. Same axis the tier ladder already uses — *pick by the shape of the work, not its importance*.

## Three copies were missing, and one was in the file this PR already edits

- **`install-wrappers.sh`** printed `Do NOT preload every file` and, forty lines later, defaulted a no-task `spawn` to `load my declared + observed context`. **Both landed in the same task file** — the worker was told to do opposite things. The default no longer restates the rule the preamble carries.
- **`install-wrappers.ps1`** had neither the preamble nor the corrected default, so the change reached two platforms of three. Both added, and the floor is now asserted identical across platforms.
- **`AGENTS.md`**, the portable contract, still said read every file in both folders. Aligned.

That is your own thesis landing on the PR — *"two hand-maintained copies of one rule is exactly how a floor goes missing from one side"* — there were four, not two.

## The suite: 7 → 15

Two of your original checks were **fixed rather than worked around**, both brittle in ways `antifragile` #105 names: one was pinned to the phrase `on demand` and broke the moment the prose improved; one extracted step 4 as a single **line** when step 4 is now a block, so it measured a fraction of the rule and reported on the rest. Comments are stripped before matching, because the commit that removes a string explains itself in a comment containing it.

Three controls verified to turn the suite red: restoring the contradicting default task, stripping the `.ps1` preamble, and re-keying the rule on identity.

53/53 suites green, bash 3.2 verified. Changelog rewritten for the new shape and kept in its own `2026-09-12` block — you were right that 09-11 had no budget left.

**Not merging yet** — this is for you and the maintainer to look over first.

chuycepeda:
CI green — and the Windows job earned its keep twice on the way there. Both failures were mine, in the `.ps1` half this PR previously did not cover.

**1. Non-ASCII in a BOM-less `.ps1`.** I wrote em dashes into the preamble. Windows PowerShell 5.1 corrupts non-ASCII in a BOM-less script at parse time, and the repo guards exactly that. Replaced with `--`.

**2. A nested single-quoted here-string — the interesting one.** The fix for (1) surfaced `The '--' operator works only on variables or on properties`, in the *generated profile*. Cause: the whole `spawn` function lives inside the single-quoted here-string `$WRAPPER` (opens line 159), and I opened another one inside it. PowerShell ends a here-string at the first **line-initial** terminator, so mine closed `$WRAPPER` at line 375 instead of 493 — everything after became live code, and `--` parsed as an operator. There is no escape for that terminator inside a single-quoted here-string.

The preamble is now a **string array joined on `[Environment]::NewLine`**, deliberately apostrophe-free so no element needs escaping. The `"produces a profile that PARSES"` job is what caught it, which is precisely what that job exists for.

One assertion of mine was also fixed rather than worked around: it matched the phrase `own words, or act on their behalf`, and the `.ps1` has to say `the words of the operator` to stay apostrophe-free. A check pinned to one file's wording fails on a purely mechanical change — the same `#105` shape as the two of yours that were fixed earlier. It now matches the rule (`act on their behalf`), with a control proving it still catches a file that drops the test.

**Also fixed: the `Unverified` badge on my commit.** Unrelated to this change — after an account switch I had been passing `-c user.email=` on raw `git commit` calls with an address GitHub cannot map to an account (`reason=no_user`), so the signature was present but unverifiable. Re-authored with the configured identity; it reads `verified=true` now.

53/53 suites green, 14/14 CI. Still **not merged** — yours and the maintainer's call.

chuycepeda:
Three more changes, all from checking the floor against a real vault rather than against the idea of one. **A correction to my earlier comment comes first, because I overstated your design.**

## Correction

I wrote that a `content-writer` would skip `personal_voice.md` under your rule. **That was wrong** — `content-writer.md` explicitly instructs itself to read `personal_voice`, `about_me` and `about_business`, so under your rule it reads them. **20 of 35 bundled agents self-declare like that**, and of the 11 that don't, 9 are engineering or visual work where your on-demand read is simply correct. Your design lands right for most of the roster, and my claim was too strong. The residual gap is narrower: agents that neither self-declare nor do checkable work (`lawyer` is the clear one), plus anything with no roster to fall back on — an operator's own `custom/` agents, and `spawn <adj-animal> "write the investor update"`.

## 1. The floor was thinner than it read

`grep '^## '` on growth + patterns returns **56 words** — ten section names like *"Work patterns"*. It reads as "load the structure of two key files" and delivers a table of contents. The index that matters is one level down: `patterns` has **73** `###` entries, `growth` 36, `preferences` 50, `antifragile` **134**.

The floor is now the **`###` entry titles** of `antifragile` + `preferences` + `patterns` + `growth` — still cheap (a few hundred lines, ~4% of those files), but actionable. **`antifragile` was missing entirely and has the strongest claim:** `CLAUDE.md` already calls it the file to *scan for relevant rules before executing commands*, and an entry title like `110. Empty and no-op inputs resolve to something REAL` tells a worker both that the lesson exists and when to open it.

## 2. What is skipped is named, and stays reachable

Operator-raised, and it is what makes a narrowed floor safe rather than a ceiling. The rule read as a **one-time boot decision** — a worker that discovered mid-task it needed the voice file had no signal that reaching for it was allowed. The preamble now names the scale it is skipping (~150k tokens), says it is skipped because the task does not need it **and not because it is off limits**, and states that opening any of it mid-task is expected. With the tell spelled out: *if you are guessing at how they would phrase something, stop guessing and read.*

## 3. The narrowing stays measured

`/aios:housekeeping` gains a report-only bucket naming any spawned worker that did real work and loaded **zero** context files — with primary sessions as its control, so a broken detector reports nothing rather than a reassuring number.

## The measurement behind all of it

Across one live vault's spawned workers with ≥20 tool calls: **3 read zero context files, 9 read one or two, 12 read three to six, 4 approached the full set.** The stated rule was "read everything", and the full ritual is ~117k words / ~150k tokens before the first question — which is exactly why it was not happening. Your diagnosis was right and the evidence for it is stronger than n=3 suggested.

Suite **15 → 23**. Worth noting your design paid off here: the floor is *derived* from the preamble rather than restated, so adding two files to it needed no test edit at all.

53/53 green, 14/14 CI. Still **not merged**.

chuycepeda:
Pushed an iteration on top of your branch (`d6c0409`). CI is green, 14/14, including the Windows wrapper installer that was red.

**Short version:** your instinct was right and the measurement backed you — but the axis was wrong. It isn't *how much* to load, it's *what kind of thing* is being loaded.

### What the measurement said

I instrumented ~28 real spawned workers on a live vault. The full ritual is ~117k words (~150k tokens). Distribution of context files actually read before doing work: **1 worker at zero, 6 at one or two, 10 at three to six, 11 at seven or more.** So "everyone reads everything" was already not true — it had quietly degraded into a TLDR, which is the strongest argument for your PR that exists. The rule was being ignored, and nobody knew.

Then a live A/B on one identical task — draft a two-sentence LinkedIn post announcing a release — across three arms:

| arm | tool calls | outcome |
|---|---|---|
| today's rule (read everything) | **0** | generic post + an honest caveat that it didn't know the voice |
| read-everything, enforced | **38** | 18 files opened, **no deliverable at all** |
| your PR's on-demand rule | **~4** | mechanical; never reached for voice |
| this iteration | **26** | post in the operator's actual voice |

The 38-call arm is the one that changed my mind about the PR. Enforcing the current rule doesn't produce a better deliverable — it produces *no* deliverable. Your PR is aimed at a real defect.

### Why the on-demand rule alone wasn't enough

The failure mode I couldn't accept was a worker that doesn't *know it doesn't know*. Voice is exactly that: a worker drafting in someone's name has no signal telling it that a voice file exists, so it writes plausible generic prose and never reaches. "Load on demand" only works if demand is detectable, and for identity it isn't.

### The split that resolved it

The two context folders are different kinds of file, and that — not volume — is what should decide how they're read:

- **`declared/`** — ~20k words of identity prose. It **does not index.** There are no entry titles to scan, so a worker either reads it or guesses at the voice. There is no cheap middle.
- **`observed/`** — ~100k words written as `###` entries. It **indexes beautifully.** A title tells you the lesson exists and when it applies, at a few hundred lines instead of a hundred thousand words.

So:

1. **The map, always, whatever the task.** Both folders listed, both `_index.md` read, plus the `###` entry titles of `antifragile.md` / `preferences.md` / `patterns.md` / `growth.md`. That's the floor *and* the index in one pass — and `antifragile.md` matters most, since the ritual already asks workers to scan it before executing commands.
2. **Then one question about the worker's own output:** *will this be read as the operator's words, or act on their behalf?* Yes → all of `declared/` + `INTENT.md`. No → the map is enough.
3. **What's skipped stays reachable, and the worker is told so in those words** — it's skipping `observed/` because titles index it better than prose does, **not because it's off limits.** Reaching mid-task is expected, not exceptional.

Point 3 is the one I'd defend hardest. A rule that narrows without saying *why* it narrowed teaches workers the rest of the vault is out of bounds, and you lose more than you saved.

### Keeping it honest

- The rule now lives in four places that are **asserted to agree** — `CLAUDE.md`, `AGENTS.md`, and both spawn wrappers. The parity suite went 7 → 30 assertions.
- `/aios:housekeeping` gains a report-only bucket naming any worker that did real work and loaded nothing.
- That detector ships as **code** (`hooks/context-load-audit.py`), not as prose a command re-implements. Worth saying why: my first version matched full paths only and missed every worker that `cd`s into the context folder first — it reported workers at zero that had read the whole floor, and I repeated that inflated number publicly before catching it. It now counts `tool_use` only, resolves `cd`-relative reads, separates floor-depth from full reads, is unaffected by compaction, and **refuses to report unless a primary session validates it** — without a control, "a worker read nothing" and "the detector is broken" look identical.

### Status

**Not merging this yet** — deliberately. It's on your branch so you can see it, argue with it, and push back on any of it. I'd particularly like your read on whether the output question (point 2) is the right gate, or whether there's a cleaner trigger for "this work will be read as the operator's own words."

Thanks for the nudge — it found a real problem that the existing rule was hiding.

chuycepeda:
Second iteration pushed (`bcf80af`). CI green, 14/14. It starts with a correction to my own last comment.

## I was wrong about why the split works

I justified reading `declared/` whole by saying that folder **"does not index"** — prose, no entry titles, read it or guess. I wrote that into `CLAUDE.md`, `AGENTS.md`, both spawn wrappers, the CHANGELOG, the comment above, **and into a test that asserted the wording was present** — so the guard enforced the claim instead of checking it.

I never measured it. Measured, it is false. Words per heading:

- `declared/` — **70–140** on a mature vault, **36** in a fresh clone
- `observed/` — **160–439**

`declared/` indexes about twice as densely as `observed/`. It is the *better*-indexed folder. Four shell commands would have caught this before it shipped.

**The rule survives; the reason is replaced.** The real one is an asymmetry in how the two failures surface:

- Under-read `observed/` → the gap **shows**. Work is visibly missing something, or you know to ask, or a title you saw nags at you.
- Under-read `declared/` → the work comes back fluent, correct, and **not theirs**, with nothing in it looking wrong.

You cannot detect the absence of a voice from inside your own output. Size is only what makes acting on that cheap. The parity suite now asserts the true reason **and** carries a negative assertion so the false one cannot come back.

## The bug the correction exposed is bigger than the PR

**"Read everything" was correct when it was written.** A fresh clone's entire operator context is a few thousand tokens — cheaper than the paragraph telling a worker to skip it. It silently stopped being correct as vaults grew past six figures, and nothing reported the crossing.

That is not a bug in the old rule. It is a bug in **having written a volume down at all**. A constant baked into a rule about a quantity that grows works, then doesn't, and nobody is told.

So no file states a volume any more. They ask:

```
$ python3 hooks/context-rungs.py
```

It reports the four rungs **in the vault in front of it** — both `_index.md` · every heading in both folders (the floor) · all of `declared/` · everything — in words and estimated tokens, ending in a verdict that **flips**:

- **fresh clone** → *"this whole context is ~7k tokens. Read ALL of it. The ladder exists for vaults that have grown past this. Yours has not."*
- **mature vault** → *"rung 3 is ~161k tokens — too much to spend before the first question. Floor at rung 1."*

Same rule, opposite advice, both right. Your PR and the old rule were each correct for a different vault size, and neither could say so.

It **refuses** rather than reporting a partial ladder — a missing folder exits 2 instead of totalling the one that is there, because a total built from one folder reads as *small*, and small is exactly the answer that talks a session out of loading anything. `tests/context-rungs.test.sh` runs the refusal cases first and carries a control that fails if a small and a grown vault ever get the same verdict (verified by mutation: pin the verdict and the control fires).

## The `right-context` skill

The ladder, the escalation tells, and the mid-task climb now live in `skills/aios/right-context/`. **The floor and the one question stay unconditional in `CLAUDE.md`** — the measured failure was workers ignoring an unenforced rule, and a floor living in a skill you must remember to load inherits that failure exactly. The parity suite guards that split in both directions.

The tells it carries are the practical part: *you are writing "probably" about their own preference* · *a person appears and you are reconstructing the relationship from the task text* · *you are about to publish, send or commit* · *a title you saw at rung 1 keeps coming to mind*. **Guessing is not a rung.**

There is a recursion here I think is load-bearing rather than cute: every word in `CLAUDE.md` is paid by every session forever, so the triage detail belongs behind a load-on-demand door. **The rule about what context to load now obeys the rule about what context to load.**

Also added: `/aios:housekeeping` **Bucket 31** runs the measurement on cadence and flags when the ratio crosses — with an explicit instruction *not* to propose restructuring `declared/` on the strength of it. That folder is the operator's own words, and it is already the denser-indexed of the two. That instruction exists because I got it wrong once.

## Status

**Still not merging.** Two things I'd genuinely like your read on:

1. Is the output question (*"will this be read as the operator's own words?"*) the right gate, or is there a cleaner trigger?
2. The verdict threshold in `context-rungs.py` is one number I picked (~25k tokens = "just read everything"). That is the one constant left in the system, and I am not confident in it. It should probably be derived from something rather than chosen.

Thanks for the nudge. The PR found a real defect; chasing *why* it was a defect found a larger one.

Issue #142, opened 2026-09-16
mcps: google-workspace connect.sh case-match fails on Windows (CRLF from python)


bernardobitardladv-hash:
## Summary

`mcps/google-workspace-mcp/connect.sh` fails to match its `case` branches on Windows because the value it matches against carries a trailing `\r`.

## Root cause

The script reads a value produced by Python. On Windows, `python`/`python3` writes `\r\n` line endings, so the captured value ends in `\r`. The subsequent `case "$g" in ...)` then compares `value<CR>` against bare patterns and never matches — the connect flow silently takes the wrong branch.

## Fix

One line, immediately before the `case`:

```sh
g="${g%$'\r'}"  # Windows python3 writes CRLF; strip stray \r before case-matching
```

Inert on macOS/Linux, where there is no `\r` to strip.

## Note on how this was found

My vault already carries this line — a previous session added it locally to get the connect flow working on Windows. I noticed it because `/aios:update`'s three-way compare correctly classified my copy as **AHEAD** (canonical had not touched the file since my last sync) and declined to overwrite it. That branch working is why the fix survived the sync instead of being silently reverted — worth noting as a datapoint that the anti-downgrade rule earns its keep.

Happy to send this as a PR if preferred.

## Environment

- Windows 11, Git Bash (MINGW64)
- Framework at `585f3d3`

chuycepeda:
Fixed and merged in cb6ed39 (#140).

The fix is the line you proposed, at the site it belongs to — `derive_apis()`'s read loop in `mcps/google-workspace-mcp/connect.sh`, immediately before `api_for()`'s `case`. Placed *before* the emptiness test on purpose: a line holding only a CR is non-empty and would otherwise be reported as an unmapped group named `""`.

One correction to the issue's description, because it makes the symptom easier to recognise: it does not take the wrong branch silently. Every mappable group falls into `unmapped`, so the run **stops** — and the error then names rows `api_for()` plainly has, because a CR is invisible in terminal output (it only returns the cursor). Reading `no Google API is mapped for permission group(s): drive docs sheets…` while looking at a `drive)` row two screens up is the tell.

Two mechanisms rather than one, since neither alone is sufficient:

[...]

Issue #141, opened 2026-09-16
update: Tier-0/Tier-2 denylist matches nothing on Windows Git Bash — reconcile reports operator context as framework drift


bernardobitardladv-hash:
## Summary

On Windows Git Bash, the Tier-0/Tier-2 denylist in `/aios:update`'s completeness reconcile (Step 6.5) matches **nothing**, so the reconcile compares `vault/` — the operator's own content — and reports their entire declared + observed context as "framework drift to pull".

Following the command's own instruction for those lines ("apply it exactly like a Step-3 Tier-1 file") would overwrite the operator's accumulated context with canonical's empty seed files.

The reconcile exits 0. The output lines are indistinguishable from legitimate drift. Nothing warns.

## Root cause

Both Step 2 and Step 6.5 derive the layer list this way:

```bash
for p in $(cd "$CLONE" && ls -A); do
  [ -d "$CLONE/$p" ] || continue
  case " $TIER0_DENY " in *" $p "*) continue ;; esac
  ...
```

**Git for Windows ships `ls` aliased to `ls -F --color=auto --show-control-chars` by default** (`/etc/profile.d/aliases.sh`). `-F` / `--classify` appends a type suffix: `/` for directories, `@` for symlinks.

So `$p` is `vault/`, `tests/`, `.git/`, `.github/` — and never `vault`, `tests`, `.git`, `.github`. Every denylist entry misses.

Verified on my machine:

```
$ type ls
ls is aliased to `ls -F --color=auto --show-control-chars'

$ for p in $(cd "$CLONE" && ls -A); do
    case "$p" in tests|.github|vault|.git|.gitattributes) echo "EXCLUDE: [$p]"; continue ;; esac
    [ -d "$CLONE/$p" ] && echo "include: [$p]"
  done
include: [.claude-plugin/]
include: [.git/]
EXCLUDE: [.gitattributes]      <-- a FILE, gets no suffix, so it matches
include: [.github/]
include: [agents/]
include: [hooks/]
include: [mcps/]
include: [plugins/]
include: [skills/]
include: [templates/]
include: [tests/]
include: [vault/]              <-- Tier 2, operator content
```

`.gitattributes` matching is what makes this look partially functional — it's a file, so it gets no classify suffix.

## Impact

The reconcile then emitted these as framework drift on my vault:

```
Files .../vault/00 - notes/context/declared/about_me.md ... differ
Files .../vault/00 - notes/context/declared/personal_voice.md ... differ
[...]

PR #146, opened 2026-09-16
fix(windows): two probes that answered the wrong question

giuliana-cmd:
Two probes that answered a question *adjacent* to the one being asked, and so returned a confident wrong result — on Windows only, silently, in both cases.

This is the class #141 named. That fix established the rule (*assert the outcome, never the code*) after `ls` was asked "what are these directories called" and answered with the caller's classify flags applied. Two instances survived the sweep.

## 1 — `command -v python3` is not a check that python3 works

Windows installs a Microsoft Store **App Execution Alias** at exactly `python3`. It is a real file on PATH, so every existence probe succeeds. Invoking it exits **49**, prints a localized *"install from the Store"* notice, and produces no stdout.

Measured on Git Bash:

```
$ command -v python3
/c/Users/.../AppData/Local/Microsoft/WindowsApps/python3     # exit 0 — the guard passes
$ python3 -c 'print(1)'                                      # exit 49, no stdout
```

Consequences:

- **`mcps/google-workspace-mcp/connect.sh`** — `need_python()` used exactly that probe, so the script **passed its own precondition** and then read empty JSON from every heredoc. It surfaces to the operator as a `connector.json` problem it is not. Note this is the script #142 just fixed for a Windows CRLF symptom; the symptom was downstream of a script that could not run on Windows at all.
- **`/aios:update` Step 3 cache-parity** — the same call carries `2>/dev/null`, so the failure is *completely* invisible. `MANIFEST` comes back **empty** and the parity test compares every cached version against `""`, reporting a mismatch and prescribing a re-resolve that realigns nothing, on every Windows sync that touches a command file. The spec already guards an empty `$CACHED`; it never guarded an empty `$MANIFEST`. An unresolved manifest is now treated as a measurement that did not complete — which is this command's own stated rule about failed measurements.
- **`tests/update-changelog-scan.test.sh`** — failed **three assertions** on Windows for reasons unrelated to changelogs, including a control that announces `CONTROL FAILED — got []; assertion 14 proves nothing`. A control failing for an environmental reason is worse than a missing one: the suite reports a real problem and points at the wrong thing. Now 13 passed / 0 failed where it was 10 / 3.

**The fix already existed in this repo.** `hooks/claude-identity/claude-identity.sh` and `mcps/setup.sh` both probe `python3 · python · py -3` and require one to actually run — with a comment naming this exact stub. It had simply never reached these three sites. This PR adopts that probe verbatim in shape, including `$PY` used unquoted so `py -3` word-splits.

## 2 — `diff -rq` returned a content verdict without normalizing line endings

§ Backup-on-divergence states the invariant: *"Every content comparison in this command strips `\r` before hashing."* The hash helpers (`h_file`/`h_git`) honour it. The **four `diff` sites** — Step 2.7's dual-owned compare and Step 6.5's reconcile — never did. *Hashing* was the only mechanism the sentence named, which is how a half-true invariant read as complete for two months.

With `core.autocrlf` the vault checks out CRLF while the clone is LF, so **every framework file reads as drift**. Measured on a live Windows vault during a real `/aios:update`:

| | `differ` lines |
|---|---|
| `diff -rq` as shipped | **267** |
| same run, CRLF-normalized | **1** — and that one was real (a genuinely AHEAD file) |

Step 6.5 instructs a session to apply each such line like a Tier-1 file, and Step 7 gates the tracker on the reconcile coming back clean — which on Windows it never could.

`--strip-trailing-cr` is **probed, not assumed**: the reconcile builds a known CRLF/LF pair and requires the flag to actually equate them before using it. #143 already traded a Windows bug for a macOS one by trusting a GNU-only flag (`find -printf`), so this deliberately does not repeat that. A diff without the flag predates CRLF checkouts being a concern, so on that platform the un-normalized compare is the correct one and the run proceeds; the one combination that cannot be true — Windows *and* no flag — says so rather than emitting a drift list nobody should act on.

## Test

`tests/update-windows-probes.test.sh` **reproduces both conditions** rather than grepping for reassuring strings — the shape `update-tier0-denylist.test.sh` argues for. It builds a stub that satisfies `command -v` and exits 49, asserts the **old guard passes on it** (the control that fails loudly if the condition ever stops reproducing), and requires the resolver to skip it. Same for CRLF.

```
against 0fd0fb7 (unfixed):   5 passed, 4 failed
against this branch:         9 passed, 0 failed
```

Existing `update-*` tests stay green.

## Checklist

- [x] Right home (canonical — the bugs are in shipped infra)
- [x] Zero operator data in the diff (grepped)
- [x] `CHANGELOG.md` entry added, `hash:` line present and empty
- [x] Test added that cannot pass vacuously
- [ ] Indexes/counts — none affected (no new command or agent)

## Not in this PR

Roughly half the test suite fails on Windows, and **31 test files** invoke `python3` in a command position. That is the same root cause at a much larger scale and deserves its own PR rather than inflating this one.

https://claude.ai/code/session_01Mex5F4oJDB3YjDQsmAUhdX


chuycepeda:
Rebased onto `main` (now carrying #157), **dropped the CHANGELOG hunk** from the branch, and **fixed the two red checks myself** rather than bouncing them back.

Both failures were our own guards working correctly, and neither was a defect in your change:

- `tests/update-windows-probes.test.sh` was committed but invoked by no workflow — *"a suite nothing runs is worse than no suite: it reads as coverage."* Now wired into `validate.yml` beside the sibling `update-*` suites, with the reasoning in the step comment.
- The same file was tracked `100644`. Now `100755`.

Those are the two steps a contributor cannot discover, because nothing in this repo tells them — no PR template, no checklist. That is a gap on our side, and this PR is now the evidence for the contribution-intake key that fixes it.

On the change itself: the CRLF half is the one that made me sit up. **267 `differ` lines, 266 of them false**, on a reconcile whose output Step 6.5 tells a session to *apply like a Tier-1 file* — so the failure mode was a Windows operator being instructed to overwrite their own tree. And you were right to probe `--strip-trailing-cr` rather than assume it: #143 traded a Windows bug for a macOS one by trusting GNU-only `find -printf`, and not repeating that was the correct read.

The CHANGELOG text isn't lost — it's the raw material for the consolidated `2026-09-22` entry, and the proof stays here in the PR body where our own contract says it belongs.

PR #147, opened 2026-09-16
fix(tests): the suite could not run on Windows — python stub and HOME isolation

giuliana-cmd:
Follow-up to #146, which fixed the same root cause where it reached *operators*. This fixes it where it reaches *contributors*: on Windows the test suite largely could not run, so a Windows contributor cannot validate their own work.

Measured on Git Bash: **34 of 61 tests failed. 29 of those 34 shell out to `python3`.**

## 1 — `python3` is a Store stub on Windows

A Microsoft Store **App Execution Alias** occupies that exact name: a real file on PATH, so every existence probe succeeds, while invoking it exits **49** and produces no stdout. A test that shells out to it does not fail for its own reason.

The sharpest example is `update-changelog-scan.test.sh` (fixed in #146), which reported:

```
FAIL 14. could not locate the newest dated entry
FAIL 15. CONTROL FAILED — got []; assertion 14 proves nothing
```

A control failing for an *environmental* reason is worse than a missing control: the suite announces a real problem and points at the wrong thing.

**28 test files** now resolve an interpreter that actually runs — `python3 · python · py -3`, requiring one to execute — and **SKIP loudly** when none is found instead of failing obscurely. This is the probe `hooks/claude-identity/claude-identity.sh` and `mcps/setup.sh` already use, with a comment naming this same stub. `$PYBIN` is used unquoted so `py -3` word-splits.

## 2 — `HOME=` does not isolate a Python hook on Windows

This one is worse than a failed test. Python's `Path.home()` / `expanduser('~')` read **`USERPROFILE`** and ignore `HOME`:

```
$ python -c "import os,pathlib; os.environ['HOME']='/fake'; print(pathlib.Path.home())"
C:\Users\<real-user>          # the fixture is ignored
```

So a test isolating a hook with `HOME="$TMP"` silently pointed it at the operator's **real `~/.claude`** — the result then depends on their machine rather than the fixture. Seven tests isolate this way; the four that invoke Python hooks now set `USERPROFILE` alongside `HOME`. A no-op on macOS/Linux.

`context-load-audit.test.sh` is the clean demonstration: it was scoring `-1` on every assertion because the hook, pointed at the real home, found no matching session and returned `[]`. It passes now.

## Result

| | Windows |
|---|---|
| before | 34 failing / 61 |
| after | **19 failing / 61** |

The remaining 19 fail for unrelated reasons — POSIX file-mode assertions (`0600`) that do not map to Windows, and content assertions — and are deliberately left untouched. They are a separate question and should not ride in on a portability fix.

**No assertion was weakened and no expectation changed.** Only how the suite reaches an interpreter, and how it scopes a fake home.

## Checklist

- [x] Right home (canonical — `tests/` is contributor infra)
- [x] Zero operator data in the diff (grepped)
- [x] No CHANGELOG entry — Tier-0 (`tests/` never reaches an operator vault), and #146 already carries the operator-facing entry for this class
- [x] Verified by running the suite before and after on the affected platform

https://claude.ai/code/session_01Mex5F4oJDB3YjDQsmAUhdX


chuycepeda:
Rebased onto `main`, which now carries #148 — so this landed on top of the multi-account work rather than beside it, deliberately: the mechanical sweep adapts to the substantive change, not the reverse.

One real conflict, in `tests/quota-rotation-target.test.sh`, where #148 made `seed_state` build a **captured** account (a rotation target must be restorable) and this PR swapped `python3` for `$PYBIN`. Both are wanted, so the resolution keeps both:

```sh
  # A rotation target must be restorable, so the fixture is a CAPTURED account.
  for f in keychain.json oauthAccount.json userID.txt; do echo '{}' > "$WORK/.claude/identities/$1/$f"; done
  $PYBIN - "$WORK/.claude/identities/$1/last-limits.json" "$2" "$3" <<'PYEOF'
```

My first pass at that resolution ate the newline after the heredoc marker and the suite refused to parse — caught by running it, not by reading it. Fixed and re-verified: `quota-rotation-target` 10/10, full suite 66/67 locally (the one red is this machine's unaccepted Xcode license, green in CI).

The finding I keep coming back to is **34 of 61 failing, 29 of them on the same Store stub**. A Windows contributor could not validate their own work, which means every Windows fix before this one was shipped on someone else's word. `HOME=` not isolating a Python hook because `Path.home()` reads `USERPROFILE` is the sharper half — that is a fixture silently pointing at a real home directory, which is worse than a failed test.

No CHANGELOG hunk on this branch, so nothing to fold; `tests/` is Tier 0 and never reaches an operator vault anyway. It gets one line in the consolidated `2026-09-22` entry for contributors.

PR #148, opened 2026-09-18
fix(claude-identity): multi-account rotation — token sessions, one switch at a time, current samples only

matiasmacera:
Quota rotation across several accounts had a family of silent failures: nothing errored, and the autopilot either rotated on the wrong numbers or lost an account's saved login. This PR closes the ones we could find in `hooks/claude-identity/`.

## 1. Token sessions stay out of the seat's rotation

A session launched with `CLAUDE_CODE_OAUTH_TOKEN` runs on the token's account, not on the *seat* (the account `~/.claude.json` names, the one rotation swaps). It still reported its limits into the seat's shared cache, which could hide the seat's cap, rotate the seat for a cap it did not have, or file the other account's usage under the seat.

- `_cache.py`: a token session writes its limits to `identities/<AIOS_ACCOUNT_EMAIL>/last-limits.json` and never touches the seat cache or kicks the watcher. With no `AIOS_ACCOUNT_EMAIL` the sample is dropped. The variable alone (no token) is ignored.
- `context-monitor.py`: the chip shows the token's account (or `👤 token?`), and there is no seat-swap banner in token sessions. `whoami` says when the session runs on a token.
- `_watch.py`: only **restorable** accounts are rotation targets. An account only ever observed used to be picked, and `switch` then failed on every tick. An uncaptured-only pool logs and waits instead of alerting every tick.
- `list` marks `✓ saved` only with all three capture files. The setup guide detects capture by files, not by folder.

## 2. One switch at a time, on the current seat only

**The main bug.** Overlapping switches (two watcher ticks, the fallback agent, a manual `claude-switch`) both read the outgoing account. The later one saved the credential the earlier one had just installed **under the outgoing account's name**, and that account's login was gone. This reproduces deterministically against `main` (test § 6).

- **Seat lock.** `switch` and `--capture` hold an `flock` on an fd the script keeps open, so the lock is released with its holders and a crash never leaves it stale. A contender exits **75** and changes nothing. A Python without `fcntl` (native Windows) warns and runs unlocked, as before. Any other failure to lock aborts.
- **Seat generation**, bumped on every completed swap. The watcher passes the generation and seat its decision was made on (`AIOS_SWITCH_EXPECT_GEN` / `_FROM`), and `switch` refuses with 75 if either moved, including A→B→A. A 75 or `already on` writes no swap row, sends no notification and arms no cooldown.
- **Swap integrity.**
  - `restore_identity` validates the target before touching the seat and writes `.claude.json` via `mkstemp`.
  - If the metadata write fails, it judges by what actually landed and otherwise rolls the credential back.
  - A pending marker spans the two writes. If a swap dies halfway, the next switch finishes it **without capturing the half-written seat** as the outgoing account, and `--capture` refuses until then.
  - A failed recovery keeps the marker. `already on` is only answered for a consistent seat.
- **Only current samples.** The watcher skips a sample that:
  - is from another generation;
  - names another account than the seat;
  - is older than the last logged swap;
  - was taken with an unreadable seat.

  Nothing is recorded inside the cooldown. `_cache.py` records the generation **before** reading the payload. Manual swaps are logged.
- **Whole files only.** One atomic writer (`_fs.py`, `mkstemp` + `os.replace`) for the cache, per-account limits, the rotation alert and the swap marker. A fixed `.tmp` shared by concurrent writers could publish torn JSON, and an unreadable `last-limits.json` reads as "fresh", which turns a capped account into a target.
- **Pause file.** The watcher no longer deletes it. Read-then-delete could remove a new pause written in between. An expired or malformed marker is ignored, and an empty one seconds old (being written) skips the tick. `tests/quota-watch-pause.test.sh` is updated to this contract.

**Stated limit, in the README.** The statusLine payload does not say which account served the request its numbers came from. The generation catches every swap that lands while a sample is being written. A session still showing its last pre-swap reply is covered only by the cooldown.

## Verification

- `tests/token-session-account.test.sh`: 23 checks.
- `tests/multi-account-consistency.test.sh`: 32 checks. Interleavings are forced with a `cat` shim and FIFOs, not timing, and every wait has a deadline, so a broken lock fails instead of hanging. File backend via a `uname` shim, so no Keychain is touched.
- `hooks/claude-identity/test-identity-cycle.sh`: cases 8–12 added (39 checks). The Keychain cases use the suite's isolated service and clean up their lock files.
- **Against `main`'s code**, the new consistency suite fails 20 of 32. Targeted reverts of individual fixes each fail their own case.
- Full `tests/` directory plus the identity cycle pass locally under bash 5 and macOS `/bin/bash` 3.2.

https://claude.ai/code/session_011WxdF3ydZMUNoJHZPYsanm


chuycepeda:
Rebased onto `main` (now carrying #157) and **dropped the CHANGELOG hunk from this branch** — not because the entry was wrong, but because five PRs are landing in one release today and canonical runs one entry per day. Your text is the raw material for the consolidated `2026-09-22` entry; the proof stays here in the PR body, which is where our own CHANGELOG contract says it belongs.

Nothing else on the branch was touched. Verified after the rebase, locally: `token-session-account` 23/23 · `multi-account-consistency` 32/32 · `quota-rotation-target` 10/10 · `quota-watch-pause` 14/14, alongside #157's suites.

The seat-lock + generation design is the part that sold this — an overlapping switch saving the incoming credential under the outgoing account's name is exactly the silent class we keep finding, and reproducing it deterministically against `main` is what made it reviewable rather than plausible.

PR #149, opened 2026-09-19
fix(pipeline): render calendar times in the operator's timezone

llacunaj-source:
## The problem

Google returns two different shapes for an event's `dateTime`:

| Calendar kind | What the API returns |
|---|---|
| One the account owns | `2026-09-22T15:00:00+02:00` — local offset |
| One **subscribed** from an ICS feed (`…@import.calendar.google.com`) | `2026-09-22T13:00:00Z` — normalised to UTC |

Google normalises a subscribed feed to UTC because it is not the account's to place. Both encode the same correct instant.

`google_calendar_events()` rendered times by slicing the string — `dateTime[:16]` — which reads the UTC instant as though it were already local. Every event from a subscribed calendar therefore rendered **shifted by the account's own offset**.

An Outlook / Microsoft 365 mirror is the common way to end up with one, so this hits any operator who replicates a work calendar into Google — which `calendars_skip` cannot address, because the calendar is wanted, not unwanted.

### Measured on a live vault (`Europe/Madrid`, one Monday)

| Event | Rendered | Actual |
|---|---|---|
| 1:1 (mirrored) | 07:30 | **09:30** |
| Follow-up (mirrored) | 09:00 | **11:00** |
| Closing (mirrored) | 10:10 | **12:10** |
| Training (mirrored) | 12:00 | **14:00** |
| Leadership meeting (mirrored) | 15:00 | **17:00** |
| Sports (Google-owned) | 16:00 | 16:00 ✓ |

**What makes it worth fixing structurally rather than documenting:** it fails silently *and selectively*. The Google-owned calendars stay correct on the same page, so the day's plan is internally consistent and nothing about the output says it is wrong. The operator who reported this had the plan render a full working week two hours early without a single visible symptom.

Sorting carried the same defect: `_sort_key` compared raw strings, so mixing `Z` and `+02:00` sorts by wall-clock digits. An event just after midnight local is stored as the previous evening in UTC and was filed under the wrong day.

## The fix

- `_event_instant()` parses to a real tz-aware instant and converts to the operator's timezone.
- `_local_hhmm()` renders `HH:MM` from it, falling back to the previous slicing when a value cannot be parsed — one malformed event cannot take the day's plan down.
- `_sort_key()` orders by the local instant, so the day boundary is decided correctly.
- The timezone comes from `sources["timezone"]`, already parsed from `USER.md` → `## Sources` → `### General`. No new configuration.

`timezone_name` defaults to `"UTC"`, matching the existing config default, and is threaded through all three call sites.

## Tests

`tests/calendar-timezone.test.sh` — 11 checks covering: a `Z` instant rendering local in summer and winter; an offset instant left untouched; the mirror/native pair agreeing on one instant; a foreign `TZID` field not steering the rendering; the day boundary following the local instant; malformed, empty and unknown-timezone inputs degrading rather than raising; and two guards that the renderer does not return to slicing.

**Control run:** executed against the pre-fix renderer, the probe aborts and the suite exits 1. The first version of this test reported `0 passed, 0 failed` and exit 0 in that situation — a gate certifying the bug it exists to catch — so the harness now treats the probe's own exit status as load-bearing and fails when no checks ran.

Full suite: 62/62 green on macOS.

## Notes for the maintainer

- The `timeZone` field on an event from a subscribed calendar carries the **source's raw TZID** (`Africa/Ceuta`, `Europe/Amsterdam`, `America/New_York` were all present on one account). It is not a reliable rendering hint and the fix deliberately ignores it — a New York TZID on a correctly-stored instant would otherwise become a six-hour error. There is a test for this.
- `hash:` is left empty in the CHANGELOG entry, per CONTRIBUTING.
- Happy to adjust the CHANGELOG wording — it is written for the operator who mirrors Outlook and has been quietly reading wrong times.


chuycepeda:
Rebased onto `main` (now carrying #157 and #148), **dropped the CHANGELOG hunk**, and **fixed the red check myself**: `tests/calendar-timezone.test.sh` was committed but invoked by no workflow, and was tracked `100644`. Both are now done — wired into `validate.yml` next to the sibling `pipeline-executor-*` suites, and `100755`.

You are the third contributor today to hit exactly those two steps. That is not three mistakes, it is a missing PR checklist on our side, and it is now the evidence for the key that fixes it. Nothing for you to do.

On the change: the reason this one is worth fixing structurally rather than documenting is that it fails **selectively** — the Google-owned events stay correct on the same page, so the day's plan is internally consistent and nothing in it says it is wrong. A full working week two hours early with no visible symptom is the worst shape a bug can have in a planning surface.

The detail I want to keep: your first version of the test reported `0 passed, 0 failed` and exit 0 against the unfixed renderer — *a gate certifying the bug it exists to catch* — and you made the probe's own exit status load-bearing rather than shipping it. That is precisely the discipline this repo tries to hold, arriving from outside it.

The CHANGELOG text is folded into the consolidated `2026-09-22` release entry; the proof stays in the PR body.

